PR #67, opened 2022-05-31
dynamic dropdown plus styling

JamesKunstle:
changed the dropdown bar from being fully pre-loaded and extremely slow to being dynamically filled in by the user's input, solving a large bottleneck that we've had for a while causing a 12 second slowdown upon startup. The slowness came from sending ~10,000 entry list of strings to the dropdown bar.


cdolfi (commented):
All looks good to me!

PR #87, opened 2022-06-21
Merge pre-commit local and on-push / on-pr checks into upstream

JamesKunstle:
Added pre-commit checks to project, this PR includes the configuration necessary to add pre-commit and a change to the README to reflect the state of the branch w.r.t it's pre-commit checks on-push.


JamesKunstle:
pre-commit check will fail this time because this commit only includes the configuration for pre-commit- its doesn't actually enforce the necessary formatting changes that will come in a subsequent PR.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,59 @@
+---
```
MichaelClifford, line 1: Looks good to me. Here is an example pre-commit-config we've used in the past, if you want to add some additional checks later on. https://github.com/aicoe-aiops/ocp-ci-analysis/blob/master/.pre-commit-config.yaml

but for now I think this looks good.

PR #95, opened 2022-06-23
added pipenv module resolution to Dockerfile

JamesKunstle:


requirements.txt
```diff
@@ -1,11 +1,11 @@
-dash
-dash_labs
+numpy
```
MichaelClifford, line 1: Do you still need this file if you are managing your dependencies with Pipfiles now?

JamesKunstle: very fair point

PR #109, opened 2022-06-30
Switched app to use logging instead of print

JamesKunstle:
Normal runtime info is logging.debug, real problems
are logging.error or logging.critical. Will investigate
longer-term logging in the future.

### DIFF
--- a/app.py
+++ b/app.py
@@ -6,9 +6,12 @@ import dash_labs as dl
 import dash_bootstrap_components as dbc
 import numpy as np
 from dash_bootstrap_templates import load_figure_template
+import logging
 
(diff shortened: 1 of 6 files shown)


cdolfi:
@JamesKunstle Not too familiar with logging. With the change from print statements how does this show up in the terminal working locally? On the hosted app, where would this show up?

JamesKunstle:
@cdolfi Shows up in the terminal with print statements but with the additional context of importance level- critical and error are difference from normal debug and info prints. They are printed to standard out but we use logging because it has a convenient interface to also output to a log file when we want it to in the future.

PR #110, opened 2022-07-05
Added sample config.json

sgoggins:


JamesKunstle:
merging without signature because it's new policy

PR #126, opened 2022-07-12
file reorganization

cdolfi:
This PR changes the file structure where each visualization has its own file that includes its card and layout information and the callbacks associated


JamesKunstle:
@cdolfi There are some functions (e.g. get_graph_time_values) that are shared between pages. That function should only be written once and referenced elsewhere. This might be a good use of a /utils/graph_utils.py file for this kind of logic. We shouldn't merge this PR until copied code is removed.

cdolfi:
@JamesKunstle Updates made, let me know if any other changes are needed

PR #128, opened 2022-07-14
s2i updates

codekow:


codekow:
You will need to update your OCP deployments to use the new ENV vars.

Issue #134, opened 2022-07-15
Deploy app on prod w/ gunicorn instead of Flask simple server


JamesKunstle:
Dash suggests that the app not go to production using a bare Flask server; recommend using gunicorn instead.

JamesKunstle:
Task 1: what does Gunicorn do? Why does the Dash documentation suggest that we use it?

JamesKunstle:
Figured out how to use it and what the benefit of a WSGI interface is for a Flask app. Switched to using it, now want to test how it performs when deployed.

JamesKunstle:
Solved by #136

PR #136, opened 2022-07-18
Add gunicorn

JamesKunstle:
Deploy application using gunicorn rather than raw Flask server. The dash docs recommend this practice for production-level deployments.

Solves #134


cdolfi:
tagging @codekow to see what he thinks. lgtm but I also do not have much context on gunicorn

PR #146, opened 2022-07-22
Fixes search bar clobbering

### LINKED ISSUE #71: Search bar clobbering
**Describe the bug**
When searching for repos, if one option is already selected an done begins to type another, the first selected is clobbered and removed.

JamesKunstle:
A value selected in the search bar must also be in the options
    list, otherwise it's removed.
    
    This was introducing a bug because longer repo addresses were
    being dropped but org addresses weren't.
    
    This was because for short inputs of only a few letters, the addresses
    of repos were of index >250 so they were dropped when we limited
    the length of the returnable options, which had the side-effect of removing
    them from the list.
    
    This change ensures that those values currently selected are always appended
    to the end of the options list so they won't be removed randomly.

Fixes #71

### DIFF
--- a/app.py
+++ b/app.py
@@ -155,8 +155,6 @@ app.layout = dbc.Container(
                                             is_open=False,
                                             color="info",
                                         ),
-                                    ]
-                                        )
                                     ],
                                     style={
                                         "width": "50%",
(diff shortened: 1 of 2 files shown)


cdolfi (approved):
lgtm~

cdolfi:
@JamesKunstle can you handle the merge conflict?

JamesKunstle:
@cdolfi def

JamesKunstle:
@cdolfi done

PR #147, opened 2022-07-22
Adds notice for search-bar help

JamesKunstle:
This change provides the user with a note that if they can't see their
selection in the search bar, they need to first check their spelling,
and otherwise to request that the selection be loaded in manually.


cdolfi (approved):
lgtm!

PR #151, opened 2022-08-24
PR staleness

cdolfi:
This visualization shows how long pull requests have been open and the trends over time

### DIFF
--- a/pages/visualizations/overview/pr_staleness.py
+++ b/pages/visualizations/overview/pr_staleness.py
@@ -246,6 +246,9 @@ def get_new_drifting_stale_up_to(df, date, drift_interval, stale_interval):
     # drop rows that have been closed before date
     df_lim = df_lim_created[df_lim_created["closed"] > date]
 
+    # include rows that have a null closed value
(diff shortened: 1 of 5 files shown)


pages/visualizations/overview/pr_staleness.py
```diff
@@ -0,0 +1,269 @@
...
+    if drift_interval > stale_interval:
+        return dash.no_update, True
+
+    if drift_interval is None or stale_interval is None:
+        return dash.no_update, dash.no_update
+
+    df = pd.DataFrame(df)
+
+    # change all to datetime
+    df["created"] = pd.to_datetime(df["created"], utc=True)
```
JamesKunstle, line 173: This step appears to already be done in the pr_query

JamesKunstle (changes requested):
One suggestion: let's find a different verb together to replace "Drifting" / "Drift". I think that works for the Contributors, because a person can be "Drifting Away," but there should be something better to indicate that a PR or an issue is not quite stale but has been neglected for a while.

PR #157, opened 2022-09-01
Worker queue

JamesKunstle:
major overhaul of query logic


cdolfi:
@JamesKunstle With such a major overhaul, what specifically should I be looking for here? (obv will look through everything, just want a little direction)

JamesKunstle:
@cdolfi Well realistically my highest priority is that you can understand the system that's implemented. Please read the documentation in the /job_manager folder to understand the architecture and the motivation for that architecture, and then see if what I implemented aligns with it.

PR #158, opened 2022-09-09
staleness visualizations

cdolfi:
This has the combined visualization code that is compatible with the new job queuing system

### DIFF
--- a/pages/visualizations/overview/issue_staleness.py
+++ b/pages/visualizations/overview/issue_staleness.py
@@ -134,7 +134,7 @@ gc_issue_staleness = dbc.Card(
                         ),
                         dbc.Alert(
                             children="Please ensure that 'Days Until Staling' is less than 'Days Until Stale'",
-                            id="i_staling_stale_check_alert",
+                            id="issue_staling_stale_check_alert",
                             dismissable=True,
                             fade=False,
                             is_open=False,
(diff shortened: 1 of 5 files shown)


pages/visualizations/overview/issue_staleness.py
```diff
@@ -0,0 +1,289 @@
...
+                                    ),
+                                    className="me-2",
+                                    width=2,
+                                ),
+                            ],
+                            align="center",
+                        ),
+                        dbc.Alert(
+                            children="Please ensure that 'Days Until Staling' is less than 'Days Until Stale'",
+                            id="i_staling_stale_check_alert",
```
JamesKunstle, line 137: id seems to be misspelled- confirm?

cdolfi: Mispelled how so? Looks good to me but I might be missing something

JamesKunstle (approved):
made one comment on an id's name being wrong but the logic and formatting are good

pages/visualizations/overview/issue_staleness.py
JamesKunstle: the first word in the id is 'i' instead of what I would assume, 'is'

cdolfi: the i in the front is to distinguish between the pr one. If the id is the same it will cause issues

JamesKunstle: oh 'i' for 'issue' not 'is'. could you please make that 'issue' and 'pr' then for clarity across the two alerts? that's my only comment on quality.

cdolfi: fixed!

cdolfi:
@JamesKunstle made the change

JamesKunstle (approved):
lgtm

Issue #164, opened 2022-09-26
Docker Container Errors on Ubuntu 20.04


sgoggins:
**Describe the bug**
This message repeats constantly when trying to start Docker container with `bash scripts/run_docker_container.sh`

```
[2022-09-26 16:18:45 +0000] [3401] [INFO] Worker exiting (pid: 3401)
[2022-09-26 16:18:45 +0000] [3593] [INFO] Booting worker with pid: 3593
[2022-09-26 16:18:45 +0000] [3657] [INFO] Booting worker with pid: 3657
CRITICAL:root:Required environment variable "user" not available.
CRITICAL:root:Could not get engine; check config or try later
[2022-09-26 16:18:45 +0000] [3529] [INFO] Worker exiting (pid: 3529)
[2022-09-26 16:18:45 +0000] [3721] [INFO] Booting worker with pid: 3721
[2022-09-26 16:18:45 +0000] [3785] [INFO] Booting worker with pid: 3785
CRITICAL:root:Required environment variable "user" not available.
CRITICAL:root:Could not get engine; check config or try later
[2022-09-26 16:18:46 +0000] [3593] [INFO] Worker exiting (pid: 3593)
CRITICAL:root:Required environment variable "user" not available.
CRITICAL:root:Could not get engine; check config or try later
[2022-09-26 16:18:46 +0000] [3657] [INFO] Worker exiting (pid: 3657)
[2022-09-26 16:18:46 +0000] [3849] [INFO] Booting worker with pid: 3849
CRITICAL:root:Required environment variable "user" not available.
CRITICAL:root:Could not get engine; check config or try later
[2022-09-26 16:18:46 +0000] [3721] [INFO] Worker exiting (pid: 3721)
^C[2022-09-26 16:18:46 +0000] [7] [INFO] Handling signal: int
[2022-09-26 16:18:46 +0000] [7] [WARNING] Worker with pid 3657 was terminated due to signal 2
[2022-09-26 16:18:46 +0000] [7] [WARNING] Worker with pid 3721 was terminated due to signal 2
[2022-09-26 16:18:46 +0000] [3785] [INFO] Worker exiting (pid: 3785)
[2022-09-26 16:18:46 +0000] [3849] [INFO] Worker exiting (pid: 3849)
[2022-09-26 16:18:46 +0000] [7] [INFO] Shutting down: Master
```

**To Reproduce**
1. Follow Docker installation and running instructions on Ubuntu 20.04
[...]

PR #165, opened 2022-09-26
Docker Trick

sgoggins:
I'm sure this is not the right solution, but I was able to hack the issue #164 by adding database credentials to an `env.list` file in the root directory, and modifying the scripts/*start.sh file to use those environment variables. 

Reject this and give instructions or modify so the config.json file is read. :) 

Signed-off-by: Sean Goggins <outdoors@acm.org>

### DIFF
--- a/env.list
+++ b/env.list
@@ -1,0 +1,7 @@
+USER=<<user>>
+user=<<user>>
+password=<<db password>>
+host=<<hostname>>
+port=5432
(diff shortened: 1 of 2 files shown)


JamesKunstle:
pipelined to be fixed in #166

cdolfi:
closed as changes were integrated into another pr

PR #166, opened 2022-09-27
add development dockerfiles + script

JamesKunstle:
updates development dockerfile to launch:

redis, job_queue workers, and web server

from a single script. requires only that user
supplies env.list, which is a key-value pairing of the environment variables usually set in production.

update required because it's a hassle for users to source their environment variables and to spin up however-many required worker processes manually.

### DIFF
--- a/Dockerfile.supervisor_workers
+++ b/Dockerfile.supervisor_workers
@@ -9,6 +9,12 @@ EXPOSE 8050
 # install pipenv
 RUN pip install pipenv
 
+# install supervisor
+RUN pip install supervisor 
+
+# copy config
+ADD supervisord.conf /etc/supervisor/conf.d/supervisord.conf
+
 # create a working directory
 RUN mkdir explorer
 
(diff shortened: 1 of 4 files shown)


Dockerfile.server
```diff
@@ -1,11 +1,10 @@
-FROM python:3.8.12-bullseye
```
cdolfi, line 1: whats the difference between bullseye and not?

Dockerfile.supervisor_workers
```diff
@@ -0,0 +1,39 @@
+FROM python:3.8.12
+# If you are deploying this container yourself,
+# include your own label. Example below.
+#LABEL James Kunstle <jkunstle@redhat.com>
```
cdolfi, line 4: Is this still valid?

Dockerfile.server
JamesKunstle: I *think* it has to do with the base OS, debian vs. Ubuntu etc. will check on that

Dockerfile.supervisor_workers
JamesKunstle: ... potentially not, good catch. will think about it.

JamesKunstle:
@sgoggins This will hopefully address some of your PR's questions- would love your input on how to handle the default/non-default port stuff, that's new to me because I'm mostly developing on a clean machine.

sgoggins:
@JamesKunstle : Does the pre-commit docker build failed. Is that a problem? I kind of assume it is, and it looks like it might just be a missing line.

JamesKunstle:
@sgoggins  for the moment I'm ignoring the pre-commit checks that are executed on GitHub. they're outdated because they aren't, even locally, handling our use of redis well, mypy complains even though I've solved the issue how it asks. I'm not sure why.

PR #180, opened 2022-10-06
Library docstrings

JamesKunstle:
Adds docstrings for top-level job-queue and database manager objects.

### DIFF
--- a/job_manager/job_manager.py
+++ b/job_manager/job_manager.py
@@ -10,6 +10,44 @@ import logging
 
 
 class JobManager:
+    """
+    A class that wraps the RQ worker interface (enqueuement and return retrieval).
+    Does not abstract the choices-to-be-made based on a job's status and availability.
+
+    Attributes
+    ----------
+        _redis : (private) Redis object
+            Manager for Redis in-memory cache that hosts queue for RQ workers
+            and RQ worker results.
+
+        _rq : (private) Queue object
+            Queue onto which jobs are pushed, FIFO, for RQ workers to pick up.
+
+        
+    Methods
+    -------
+        _get_job_hash(func, arglist) (private) :
+            Creates a unique hash for each job based on the job's calling
+            function and the list of repos that the function is being run with.
+
+        add_job(func, dbmc, repolist) :
+            Enqueues job onto Queue object hosted on Redis for RQ worker to pick
+            up later, keyed by job_hash.
+
+        get_job(func, arglist) :
+            Gets most updated job from Queue object, if its job_hash ID exists.
+
+        get_results(func, arglist) :
+            Returns the status of a job and the results from that job, if they exist.
+
+        job_refreshed(job) : 
+            Returns job-object refreshed with most updated contents from Queue, if
+            it exists.
+
+        get_job_status(func, arglist) :
+            Returns a job's status, if it exists.
+    
+    """
     def __init__(self):
         # Redis cache for job queue and results cache
         self._redis = Redis(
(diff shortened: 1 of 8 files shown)


JamesKunstle:
Two separate commits because two separate changes.

job_manager/job_manager.py
```diff
@@ -10,6 +10,44 @@
...
+            Manager for Redis in-memory cache that hosts queue for RQ workers
+            and RQ worker results.
+
+        _rq : (private) Queue object
+            Queue onto which jobs are pushed, FIFO, for RQ workers to pick up.
+
+        
+    Methods
+    -------
+        _get_job_hash(func, arglist) (private) :
```
cdolfi, line 29: wdym when you list things as private?

pages/utils/graph_utils.py
```diff
@@ -1,21 +1,32 @@
...
     period = "Y"
 
     # graph input values based on date interval selection
-    if interval == 86400000:  # if statement for days
-        x_r = [str(today - dt.timedelta(weeks=4)), str(today)]
-        x_name = "Day"
-        hover = "Day: %{x|%b %d, %Y}"
-        period = "D"
-    elif interval == "D":  # if statement for days
+    if interval == 86400000 or interval == "D":  # if statement for days
```
cdolfi, line 29: Did you test if this works? When I did it like this it was not producing the right results in all cases

pages/utils/graph_utils.py
```diff
@@ -25,14 +36,10 @@ def get_graph_time_values(interval):
         x_name = "Week"
         hover = "Week: %{x|%b %d, %Y}"
         period = "W"
-    elif interval == "M":  # if statement for months
-        x_r = [str(today - dt.timedelta(weeks=104)), str(today)]
-        x_name = "Month"
-        hover = "Month: %{x|%b %Y}"
-        period = "M"
-    elif interval == "M1":  # if statement for months
+    elif interval == "M" or interval == "M1":  # if statement for months
```
cdolfi, line 39: same here

queries/commits_query.py
```diff
@@ -3,14 +3,20 @@
 from db_manager.AugurInterface import AugurInterface
 
 
-def commits_query(dbmc, repo_ids):
+def commits_query(dbmc: AugurInterface, repo_ids: list(str)) -> dict:
```
cdolfi, line 6: Why is this one different than the other queries?

cdolfi (changes requested):

job_manager/job_manager.py
JamesKunstle: ah, good question. this means that it's not meant to be exposed to the user for use. python doesn't support real private/public interfaces like Java does so the underscore means that I won't necessarily want users touching that part of the method.

pages/utils/graph_utils.py
JamesKunstle: haven't tested it, I'll check to see if it seems to comply on my end.

pages/utils/graph_utils.py
JamesKunstle: (same as above)
haven't tested it, I'll check to see if it seems to comply on my end.

queries/commits_query.py
JamesKunstle: ah good question, that was the first one I updated, forgot to do this to the other queries as well. these type annotations are "nice to have"s in that they're not enforced strictly like they are in other languages, but I need to also be consistent.

pages/utils/graph_utils.py
cdolfi: @JamesKunstle will mark as resolved once its checked into

queries/commits_query.py
cdolfi: @JamesKunstle will mark as resolved once the consistency is there

pages/utils/graph_utils.py
JamesKunstle: I don't see any problems by doing this, the graphs look the same across the two implementations.

cdolfi: once the changes for the piplock is in ill test it out on my side too just to make sure. Just remember this being a niche issue and want to make 100% sure everything is good

queries/commits_query.py
cdolfi: on my side I still see (dbmc: AugurInterface, repo_ids: list(str)) -> dict: for the commits query and (dbmc, repo_ids): for the rest

JamesKunstle: It was changed in the most recent commit to this branch, it's in files changed.

JamesKunstle: checked on it, up to date.

cdolfi (approved):

Issue #161, opened 2022-09-19
Per-repo caching optimization


JamesKunstle:
Currently, we're caching user's queries by the hash of:
hash( function_name, repo_list ). 

This is sub-optimal because a delta of just one repo requires a complete re-cache of the results.

Objectives:

- [x] Cache by individual repo in the user's query
- [x] Display progress to user per-card with a progress pie-chart
- [x] optimize so that data is only collected from workers if everything is available.

Considerations for future PR's:

- [x] What if a single repo fails? should we show incomplete data?
- [ ] Replace short cache timeout w/ LRU in Redis and  24-hour to make sure data is fresh
- [x] Check OOM signal handling in worker, describe failure to user
- [ ] add refresh button to data so that cached data can be replaced

JamesKunstle:
working on this on:

https://github.com/JamesKunstle/explorer/tree/repo-caching-patch

some current considerations:

- [ ] starting work on large orgs can tie up all the workers in the queue. how can we mitigate this? more workers?

JamesKunstle:
@cdolfi for visibility

JamesKunstle:
done and linked.

PR #181, opened 2022-10-07
Repo caching patch

JamesKunstle:
per repo caching optimization
    
Switches queries and job manager logic to queueing
and caching per-repo rather than per-repo-group,
    
Adds pie-chart loading figures to graph cards,
    
removes flashing loading animations


JamesKunstle:
Solves #161

job_manager/job_manager.py
```diff
@@ -1,5 +1,5 @@
 from redis import Redis
-from rq import Queue
+from rq import Queue, Retry
```
cdolfi, line 2: whats retry?

job_manager/job_manager.py
```diff
@@ -19,9 +19,9 @@ def __init__(self):
             password=os.getenv("REDIS_PASSWORD", ""),
         )
         # RQ service connected to Redis
-        self._rq = Queue(connection=self._redis)
+        self._rq = Queue(connection=self._redis, default_timeout=600)
```
cdolfi, line 22: whats the time out for?

job_manager/job_manager.py
```diff
@@ -30,43 +30,42 @@ def _get_job_hash(self, func, arglist):
...
         # get an identifying hash of the function used and the args supplied
-        job_hash = self._get_job_hash(func, arglist)
+        job_hash = self._get_job_hash(func, repo)
 
         # grab the job from the job queue
         job = self._rq.fetch_job(job_hash)
 
         # check if job was fetched from queue
         if job is not None:
-            logging.debug(f"CACHE HIT: {func.__name__}")
```
cdolfi, line 60: Why take this out?

pages/utils/job_utils.py
```diff
@@ -36,42 +27,92 @@
     font=dict(size=18, color="orange"),
 )
 
+def _loading_graph(done, queued, retry, failed):
```
cdolfi, line 30: Can you add some doc strings/comments in?

pages/utils/job_utils.py
```diff
@@ -36,42 +27,92 @@
...
+
+    return fig
 
 def handle_job_state(jm, func, repolist):
     """
     All visualizations use this interface
     to handle whether or not the job queue / result cache
     has the data that they need.
     """
+    num_done = 0
```
cdolfi, line 52: explain the use of these

pages/utils/job_utils.py
```diff
@@ -36,42 +27,92 @@
...
+                    values=[done, queued, retry, failed]
+                )
+            ]
+        )
+    fig.update_traces(hoverinfo='label', textinfo='value', textfont_size=20,
+                    marker=dict(colors=colors, line=dict(color='#000000', width=2)))
+
+    return fig
 
 def handle_job_state(jm, func, repolist):
```
cdolfi, line 46: comments/ doc string development

pages/visualizations/overview/active_drifting_contributors.py
```diff
@@ -26,7 +26,7 @@
                     disabled=False,
                     n_intervals=1,
                     max_intervals=1,
-                    interval=800,
+                    interval=1500,
```
cdolfi, line 29: what is this interval change for?

cdolfi: (same question for all the following files)

pages/visualizations/overview/issue_staleness.py
```diff
@@ -190,9 +185,13 @@ def new_staling_issues(repolist, timer_pings, interval, staling_interval, stale_
     # create dataframe from record data
     df = pd.DataFrame(results)
 
-    # change all to datetime
-    df["created"] = pd.to_datetime(df["created"], utc=True)
-    df["closed"] = pd.to_datetime(df["closed"], utc=True)
+    try:
```
cdolfi, line 188: why these instead of testing above if the results/ df is empty

pages/visualizations/overview/pr_staleness.py
```diff
@@ -190,10 +185,13 @@ def new_staling_prs(repolist, timer_pings, interval, staling_interval, stale_int
     # create dataframe from record data
     df = pd.DataFrame(results)
 
-    # change all to datetime
```
cdolfi, line 193: same questions as above. why try all of this instead of just testing if the results/ df is empty?

scripts/launch_dev.sh
```diff
@@ -1,4 +1,9 @@
-# can pass target port number to script for redis as: bash ./this_script.sh <new_port>
+# Script builds, connects, and runs the three components of this app: Server, Worker Pool, and Redis instance.
+
+# can pass override target port number to script for redis as: bash ./this_script.sh <new_port>
+
+# if you want more/fewer worker processes in the workerpool, please edit the 'numprocs' field in supervisord.conf
+
 # TODO redis doesn't currently like this and I'm not sure why
```
cdolfi, line 7: this is slightly sus lol

scripts/launch_dev.sh
```diff
@@ -44,10 +48,10 @@ sudo docker run --rm -it --name eightknot_server \
 
 # cleanup
 printf "\nShutting down worker pool...\n"
-sudo docker stop worker_pool;
+docker stop worker_pool;
```
cdolfi, line 51: why no sudo? (on all of these)

cdolfi (changes requested):

job_manager/job_manager.py
JamesKunstle: In the 'enqueue' method below, the Retry object lets RQ schedule a task to run again if it fails. If you look in that line, there are three numbers indicating how many seconds RQ will wait from the time of failure to try again. There are three opportunities, if the job fails a fourth time it doesn't get another chance. These retrys are needed because for large sets of workers all doing queries at the same time, Augur will close the connection on some of them because it's reached its query limit.

job_manager/job_manager.py
JamesKunstle: that's how long a job is allowed to be in the 'running' state. essentially, if a worker takes longer than 600 seconds, RQ will kill it. no job should ever take that long individually (repos all download their data relatively quickly) so this is just an upper-bound that I set to overwrite the default that was smaller.

job_manager/job_manager.py
JamesKunstle: Could leave it in if preferred, it was just polluting the logs because after the first time the job is enqueued, this fires for all of the callbacks, e.g. 5x each 1.2 seconds, which was excessive.

pages/utils/job_utils.py
JamesKunstle: yep added it to the revision list.

pages/utils/job_utils.py
JamesKunstle: will add it in the documentation.

pages/visualizations/overview/active_drifting_contributors.py
JamesKunstle: so that the front-end pings the web-server less regularly. If there are lots of browsers all pinging the web-server (num_callbacks) times every 800ms that's a lot of traffic. I'm just decreasing the periodicity of that. It doesn't change the user's experience of the speediness of the app from an update perspective.

pages/visualizations/overview/issue_staleness.py
JamesKunstle: dataframe 'emptiness' is tricky, I found. A dataframe with NaN's will report False when you call 'df.empty', so it's more reliable to just catch the exception returned from trying to cast a non-existent column ('created') to a datetime object than to check for emptiness.

pages/visualizations/overview/pr_staleness.py
JamesKunstle: same answer above

scripts/launch_dev.sh
JamesKunstle: sus in a "too complicated for general users" way or sus in a "looks nefarious" sort of way?

scripts/launch_dev.sh
JamesKunstle: I learned that it's considered bad practice to have docker containers elevated to super-user privilege if not necessary. It's also uncool to make the user use 'sudo' in a script that they didn't write.

pages/utils/job_utils.py
cdolfi: @JamesKunstle sounds good, will mark as resolved once fixed

pages/utils/job_utils.py
cdolfi: @JamesKunstle sounds good, will mark as resolved once fixed

pages/utils/job_utils.py
JamesKunstle: added!

pages/utils/job_utils.py
JamesKunstle: added!

pages/visualizations/overview/issue_staleness.py
cdolfi: Would we be able to check the length/size? Just thinking for simplicity/people creating new visualizations from this structure

JamesKunstle: Yes can experiment with this in the next step, I'll make an issue for it.

scripts/launch_dev.sh
JamesKunstle: okay I'll make that issue

pages/visualizations/overview/issue_staleness.py
cdolfi: revolving with issue being opened

cdolfi (approved):

PR #183, opened 2022-10-10
Changes order of Docker steps, lowers default launch_dev perms

JamesKunstle:
Moves folder file-copy operation in Dockerfiles to step following environment installation steps so that updates to files don't trigger an environment redownload.

Also lowers default permissions in /scripts/launch_dev.sh to non-sudo. Can run everything w/ sudo permissions by:

'$ sudo bash scripts/launch_dev.sh'

### DIFF
--- a/Dockerfile.server
+++ b/Dockerfile.server
@@ -15,12 +15,16 @@ RUN mkdir explorer
 # set that directory as working dir
 WORKDIR /explorer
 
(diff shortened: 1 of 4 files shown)


cdolfi (approved):

JamesKunstle:
@cdolfi cool if we merge this and bypass the signature requirement for the time being?

PR #189, opened 2022-10-11
Fix pipfile

JamesKunstle:


JamesKunstle:
@cdolfi Quick review here- lots of whitespace fixes from pre-commit, made one change that's important in the pre-commit config that specifically ignores file 'job_manager' for mypy checks because it's mad about redis but won't allow me to fix it by installing the correct modules, so I'm just ignoring it until it sorts itself out.

cdolfi (approved):
lgtm, just add signature

cdolfi:
@JamesKunstle when signature is added ill merge

JamesKunstle:
@cdolfi Signature is on commit, Github wants that GPG signature that we agreed was unnecessary.

PR #190, opened 2022-10-11
update: logging 'pongs' added

JamesKunstle:

### DIFF
--- a/app.py
+++ b/app.py
@@ -11,10 +11,8 @@
     Having laid out the HTML-like organization of this page, we write the callbacks for this page in
     the neighbor 'app_callbacks.py' file.
 """
-from concurrent.futures import process
 import pstats
 import cProfile
-import threading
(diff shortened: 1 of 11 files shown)


cdolfi:
@JamesKunstle looking at this I still feel like it has the same issue of over saturating the logs where you can not decipher through them. If we keep the "pongs" in I think they should be very short snip bits for the meaningful logs for start and finish are noticeable. I personally think it might be good to take them out all together but if you are sold on keeping them I would condense down from "COMMITS_OVER_TIME_VIZ - PONG to "COT- PONG" for example.

JamesKunstle:
Yeah, making them shorter is cool with me. I want there to be _some_ indication that a client request has been received by the server but you're right, it's verbose at the moment.

cdolfi (approved):
lgtm

PR #194, opened 2022-10-18
updates to functions around search bar entries

cdolfi:


app.py
```diff
@@ -54,20 +56,41 @@ def _project_list_query():
...
+                    FROM
+                        repo r
+                    JOIN repo_groups rg
+                    ON rg.repo_group_id = r.repo_group_id
+                    ORDER BY rg.rg_name"""
+
+    # query for search bar entry generation
     df_search_bar = augur_db.run_query(pr_query)
 
+    # handling case sesitive options for search bar
```
JamesKunstle, line 78: misspelling: sensitive

app.py
```diff
@@ -54,20 +56,41 @@ def _project_list_query():
...
 
+    # generating search bar entries
     lower_entries = [i.lower() for i in entries]
     all_entries = list(zip(lower_entries, entries))
 
+    # generating dictionary to be used to select repo_ids from repos selected in the search bar
+    repo_dict = df_search_bar[["repo_git", "repo_id", "repo_name"]].set_index("repo_git").T.to_dict("list")
+
+    # generating dictionary to be used to get repos from selected repo groups
+    org_dict = df_search_bar.groupby("rg_name")["repo_git"].apply(list).to_dict()
```
JamesKunstle, line 91: that's smart. Could you please comment the desired outcome of these two processing steps above them? pandas chaining is difficult to follow but that process looks good.

app_callbacks.py
```diff
@@ -4,73 +4,29 @@
...
-
-        with engine.connect() as conn:
-            t = conn.execute(org_query)
-
-        results = t.all()
-        org_repo_ids = [row[0] for row in results]
-        org_repo_names = [row[1] for row in results]
+
+    # get git urls for the repos in the organization, flatten the 2D list to 1D
+    org_repos = sum([org_dict[x] for x in org_name_set], [])
```
JamesKunstle, line 24: comment the use of 'sum' here. another trick to flatten a list is the '*' operator which can be more transparent.

app_callbacks.py
```diff
@@ -4,73 +4,29 @@
...
+    # get repo values from repo dictionary
+    repo_values = [repo_dict[x] for x in org_repos]
+    # values for the repo_ids and names
+    org_repo_ids = [row[0] for row in repo_values]
+    org_repo_names = [row[1] for row in repo_values]
 
     return org_repo_ids, org_repo_names
```
JamesKunstle, line 31: this update seems very straight forward, way more optimal.

app_callbacks.py
cdolfi: What would be the way to use the * operator? Cool with whatever makes it more readable, this was just the one solution I found googling around

JamesKunstle: Here's a cool article on it, this might be more straightforward than the 'sum' trick but your call.

https://towardsdatascience.com/unpacking-operators-in-python-306ae44cd480

cdolfi: Lets keep it for now. Ill change it in a later PR when the materialized view is ready and I update that too

JamesKunstle: good w/ me

JamesKunstle (approved):
This all looks good to me

PR #196, opened 2022-10-18
Dataframe ordering

cdolfi:
With the new changes to the architecture, the data needs to be ordered in the visualization step not the query step


JamesKunstle:
@cdolfi I'll review this PR after the former one is merged because it's based on that

JamesKunstle (approved):
checked on my local machine, looks excellent.

PR #208, opened 2022-10-24
decreases default pools of workers, rq and gunicorn

JamesKunstle:
fewer workers- 4 gunicorn workers and 32 rq workers at a time

This should decrease the load on development machines.


cdolfi (approved):
lgtm

PR #209, opened 2022-10-24
updates python to v3.11

JamesKunstle:
newer python is very fast relative to 3.10.

we're fast-forwarding our versions to python's most up-to-date version to get those speed increases.

### DIFF
--- a/Dockerfile.server
+++ b/Dockerfile.server
@@ -1,4 +1,4 @@
-FROM docker.io/python:3.8.12
+FROM docker.io/python:3.10
 
 # need this for the Dash app
 EXPOSE 8050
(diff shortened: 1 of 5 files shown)


cdolfi (changes requested):
Github actions is giving a warning for not specifying version type. For dependency reasons I think it would be goo to manually give 3.11 and us continue to update it as new python versions come out

JamesKunstle:
@cdolfi okay that's fine, do you mean to say 3.11 or 3.10?

JamesKunstle:
@cdolfi 3.11 isn't stable w/ Pipenv so let's merge 3.10 and work on 3.11 later when it's more mature.

cdolfi (approved):

PR #210, opened 2022-10-25
Query optimization

JamesKunstle:
update: smaller commits query, optimized commit callback
    
The query that downloaded commit data was slow because it
had to download a lot of currently unused data.
    
Removed unused columns and added more efficient compression
to POSIX timestamps when caching data.
    
Updated commits_figure callback to be more compartmentalized
and to use a more efficient representation of the same data
with a bar-plot rather than a histogram.


JamesKunstle:
@cdolfi I've been working on callback optimization as a part of the Redis stabilization. Data model should be made much more light-weight to accomodate user requests while we move to better infrastructure.

pages/visualizations/overview/commits_over_time.py
```diff
@@ -118,39 +119,72 @@ def toggle_popover_2(n, is_open):
...
+        rangeslider_yaxis_rangemode="match",
+    )
+
+    # set the hover template
+    fig.update_traces(hovertemplate=hover + "<br>Commits: %{y}<br>")
+
+    return fig
+
+
+def process_data(results, period):
```
cdolfi, line 173: Whats the reasoning for breaking down into another function?

queries/commits_query.py
```diff
@@ -20,14 +20,19 @@ def commits_query(dbmc, repo):
     """
     logging.debug("COMMITS_DATA_QUERY - START")
 
+    # NOTE: (jkunstle) I'm removing currently unused fields from the 'SELECT' statement.
```
cdolfi, line 23: Can you add in the comments which fields you took out?

queries/commits_query.py
```diff
@@ -41,5 +46,8 @@ def commits_query(dbmc, repo):
     dbm.load_pconfig(dbmc)
     df_commits = dbm.run_query(query_string)
 
+    # massage SQL timestampdz to Posix timestamp int64- much smaller than string or datetime object.
+    df_commits["date"] = pd.to_datetime(df_commits["date"], utc=True).map(pd.Timestamp.timestamp).astype("int64")
```
cdolfi, line 50: when you get the data on the graph generation side can you use it in the same way as before?

pages/visualizations/overview/commits_over_time.py
```diff
@@ -56,14 +57,15 @@
...
                                         options=[
-                                            {
-                                                "label": "Day",
-                                                "value": 86400000,
-                                            },  # days in milliseconds for ploty use
-                                            {
-                                                "label": "Week",
-                                                "value": 604800000,
-                                            },  # weeks in milliseconds for ploty use
+                                            # TODO:
```
cdolfi, line 60: If you are going to use trend here can you change the label to trend? On my "TODOs" is to make the rest consistent on that side as well. If its trends lets call it that and same when we use daily

pages/visualizations/overview/commits_over_time.py
```diff
@@ -105,7 +107,6 @@ def toggle_popover_2(n, is_open):
     return is_open
 
 
-# callback for commits over time graph
```
cdolfi, line 108: Can you add slight comments describing each function?

pages/visualizations/overview/commits_over_time.py
cdolfi: /should this be in graph utils if its planned to be generic to be used in multiple graphs?

cdolfi (changes requested):

queries/commits_query.py
JamesKunstle: yep, you just use pd.to_datetime() on the column, you can see how it works in the callback

JamesKunstle:
UPDATE: This is purely a WIP note post for short-term visibility.

cdolfi:
@JamesKunstle can you change this to a draft if its WIP?

PR #217, opened 2022-11-08
update: use requirements.txt instead of pipenv

JamesKunstle:
pipenv is failing and shouldn't be relied upon
until we have a process to handle these failures.


cdolfi (approved):
lgtm

PR #231, opened 2022-11-30
Function break

cdolfi:
This pr breaks all visualization into multiple functions in the same format


JamesKunstle (approved):
I read through all of your changes- since you're not making logical changes and are just doing logical flow changes I'm not going through with too fine an eye. If you expect to need really close attention I'd happily re-review.

PR #234, opened 2022-12-01
Optimize commits query + change defaults

JamesKunstle:
Drops unneeded columns from data stored in Redis from commits query, 
adds 'flower' monitoring to task queue in docker-compose,
changes defaults for celery workers to be more performant.


cdolfi (approved):
lgtm

PR #246, opened 2022-12-12
file restructuring and renaming

JamesKunstle:
- moves overview, home, index, chaoss into respective pages in pages/
- deletes old files (Pipfile, Dockerfile.supervisor_workers,
  supervisord.conf)
- moves search-bar defaults into AugurManager object from app.py
- renames app_callbacks and app_layout files to index_*
- renames AugurInterface to AugurManager and AugurInterface.py to
  augur_manager.py


JamesKunstle:
@cdolfi also interested in potentially reorganizing the pages to the following structure:

8Knot
+-- pages/
 |    +-- overview/
       |     +-- overview.py
       |     +-- cards/
              |     +-- issues_card.py
              |     +-- commits_card.py
              |     +-- prs_card.py
       +-- chaoss/
       |     +-- chaoss.py
       |     +-- cards/
              |     +-- contributors_card.py
              |     +-- contributors_over_time_card.py
+-- utils/
+-- app.py
+-- app_callbacks.py
+-- app_layout.py
.
.
.

not all of the figure names are there but you get the point- it might make more sense to have each page be a folder, with the code for that page and the page's cards inside that folder together.

cdolfi (commented):
I am a HUGE fan of all of these changes. One thing I thought of is instead of having the Augurinterface.py in the db_manager, what if we came up with a file name that app_callbacks, app_layout, and Augurinterface all life? Thoughts?

JamesKunstle:
@cdolfi so would you mean that the object AugurInterface, the app_layout code, and the app_callbacks code would all be in the same .py file? I'd like to keep AugurInterface separate because it's used in the query functions but I do think that app_layout and app_callbacks should be in the same file now, and should have a different name like 'index' or 'base,' since they're the base page of the application and are now separate from the app.py file where we create and work with the celery and Dash app objects.

cdolfi:
@JamesKunstle Typo in last comment, fixed now

cdolfi:
@JamesKunstle On the cards suggestion, I personally am not a fan of using cards as the title instead of visualizations.   Also simplifying down to "issue_cards" will be too simple as there will be many visualizations that have to do with issues prs etc. Personally I think the title of the visualization files should be as close to the visualization title itself to be able to tell which is which. There is a few now that I already get confused. Im cool with moving the overview.py into the overview folder and same for chaos etc. I would support a file structure of: 

<pre>
8Knot
+-- pages/
 |     +-- overview/
 |           +-- overview.py
              |     +-- visualizations/
                |     +-- name_of_visualization.py
       +-- chaoss/
       |     +-- chaoss.py
       |     +-- visualizations/
              |     +-- name_of_visualization.py
+-- utils/
+-- app.py
+-- app_callbacks.py
+-- app_layout.py
.
.
.
</pre>

JamesKunstle:
@cdolfi okay I think that's very reasonable. What do you think about this?:

<pre>
8Knot
+-- pages/
 |     +-- overview/
              |     +-- overview.py
              |     +-- visualizations/
                     |     +-- name_of_visualization.py
       +-- chaoss/
              |     +-- chaoss.py
              |     +-- visualizations/
                    |     +-- name_of_visualization.py
     +-- index/
            |      +-- index_callbacks.py
            |      +-- index_layout.py
+-- utils/
+-- app.py
.
.
.
</pre>

cdolfi:
@JamesKunstle  Big fan of that structure, lets do it. Lets also edit the readme to detail the file structure

JamesKunstle:
@cdolfi let me know what you think!

cdolfi (changes requested):
Viz_template got deleted, I think it makes sense to be put directly in the pages directory so it is visible to contributors. Besides that looks great!

JamesKunstle:
@cdolfi got it back- accidentally deleted it w/ the visualizations directory. should be good now!

cdolfi (approved):
lgtm! Only thought is with the next PR that the readme specific in pages should be populated as well as the main readme. Maybe the way of capitalizing on this is the main readme hits high level and the pages readme goes into details

PR #260, opened 2023-01-18
navbar addition

cdolfi:


JamesKunstle (approved):
@cdolfi is logo2.png a free resource? i.e. can we use it in deployment w/o any copyright?

would also appreciate some more detailed commit messages

pages/home/visualizations/issue_metrics.py
```diff
@@ -0,0 +1,278 @@
...
+    ],
+    className="box_emissions",
+)
+
+gc_issue_metrics = dbc.Card(
+    [
+        dbc.CardBody(
+            [
+                html.H2(
+                    [html.I(className="fa-solid fa-code-commit"), "Commits"],
```
JamesKunstle, line 90: This should say "Issues"

pages/home/visualizations/pr_metrics.py
```diff
@@ -0,0 +1,410 @@
...
+    ],
+    className="box_emissions",
+)
+
+gc_pr_metrics = dbc.Card(
+    [
+        dbc.CardBody(
+            [
+                html.H2(
+                    [html.I(className="fa-solid fa-code-pull-request"), "Pull Request"],
```
JamesKunstle, line 129: This should say "Pull Requests" plural

JamesKunstle (changes requested):
couple of things I caught when I pulled your branch

cdolfi:
@JamesKunstle The image is just a place holder, do not have any intentions of actually using it but it is from a tutorial so I think it is fair game. On commit messaging definitely, my plan is to go through on commits and comments once we have a consensus that this is the design direction we want to go in

JamesKunstle:
@cdolfi If you change the card titles this'll be good to merge. I pushed some more updated changes directly in the urgency of getting a trial to Sean. Can go back and review later when we're trying to merge to dev.

cdolfi:
@JamesKunstle I am made the changes but Im still getting the error in the logs and the button change not working with the username. Can you check that out on your side?

JamesKunstle:
@cdolfi yeah will report back

PR #267, opened 2023-02-16
Feather caching

JamesKunstle:
CSV caching to pyarrow.feather
    
CSVs are very large and pandas' (de)serialization of them is slow.
    
this patch switches to using a much faster data caching strategy because feather is much faster to (de)serialize and is smaller.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -148,3 +148,5 @@ dump.rdb
 env.list
 Makefile
 run_compose.sh
+csvs_to_feather.ipynb
+debug_contrib.ipynb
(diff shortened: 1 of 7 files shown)


JamesKunstle:
@cdolfi this is a draft, fixing an issue with the contributors visualizationss

JamesKunstle:
@cdolfi okay fixed the bug, should work fine now. just have to do a 'docker system prune -af' before you try this branch because the redis 'cache' container isn't rebuilt in between composes so it's full of CSV data.

cdolfi (approved):
lgtm but im not too familiar with feather. tagging in @MichaelClifford to see if he has any thoughts

MichaelClifford:
LGTM

PR #274, opened 2023-02-22
patch home page metric failure

JamesKunstle:
metrics fail sometimes because their
connection to Augur fails. the reason for
this failure is either server-side or db-side.

in the future the metric querying should be put on the celery queue but in light of the upcoming demo I'm making this small patch to simply allow
the queries to retry if they fail with linear
retry time increases.


JamesKunstle:
closed PR because solution wasn't good.

PR #282, opened 2023-02-28
Update and rename AUGUR_INT.md to AUGUR_LOGIN.md

JamesKunstle:
updated docs of login setup


docs/AUGUR_LOGIN.md
```diff
@@ -0,0 +1,58 @@
...
+3. Log in to to Augur via 8Knot and see aggregate visualizations of their named groups.
+
+## **Setup 8Knot to enable Login and user groups**
+---
+1. Clone [github.com/oss-aspen/8Knot/dev]
+   - We use a docker compose multi-container application build strategy.
+   - Ports 8050 (*Dash*) and 6379 (*Redis*) are assumed to be available. If this isn't possible, modifications to the docker-compose.yml file may be necessary.
+   - Inter-pod name resolution is done with Docker network's DNS.
+
+2. Create an application in Augur front-end
```
cdolfi, line 18: I would specific in here that this is an "admin'"/"developer" log in. I remember the whole process was confusing to me until I understood which part was "admin'"/"developer"  and which parts was a user log in

cdolfi (changes requested):
Just the one small comment but besides that everything looks awesome

cdolfi:
@JamesKunstle bumping this

docs/AUGUR_LOGIN.md
JamesKunstle: Okay understood, will update

JamesKunstle:
ignore spelling for review, I'll fix that in another PR.

cdolfi (changes requested):
So what I meant in the prior comment about admin and developer roles (probably not the right name) is the portion of it that is admin so setting up the connection (making your account, create application toget the app_id and client secret, and then putting it into the environment. Then the portion of it that was called "dev" which a better name could be "user" describes making an account for the purpose of creating user groups and seeing them 8knot. If this doesnt make sense we can talk in our meeting about it. The most important distinction that helped my understanding was the two different natures of creating an augur login. One for connecting the app, one for user groups

JamesKunstle:
I see what you're aiming at- since these are docs that will live in the repo I'm only intending to inform the personas who are taking part in preparing the application for consumption, who I've called the "Admin" and the "Dev", or the application owner and application deployer respectively. 

I wouldn't target the end-user who creates an account and their own user-groups via the front-end in this documentation since I'm never expecting them to read it themselves. 

That said, I'll add that as a third persona in the 'Roles' section to point out that that 'Dev' or 'Admin' roles are meant to set up the instance and then the 'User' role is the consumer of the login integration having been set up.

JamesKunstle:
@cdolfi I'm separating task-roles into personas but these could certainly all be done by the same individual. Generally an application admin doesn't set an app up on bare metal like an ops person does, nor does an ops person handle the application admin perspective, so I thought it was prudent to separate those roles.

cdolfi:
@JamesKunstle cool sounds good, you def know more about this area than I do. Should I ignore the pre-commit fail as well as the spell check ones?

JamesKunstle:
@cdolfi Yeah I'm going to go back and clean that up at the repo level, we can merge this doc and then I'll fix the spelling mess.

PR #285, opened 2023-03-01
contributor activity cycle visualization

cdolfi:
This PR creates the contributor activity cycle visualization

### DIFF
--- a/pages/chaoss/visualizations/contrib_activity_cycle.py
+++ b/pages/chaoss/visualizations/contrib_activity_cycle.py
@@ -146,6 +146,7 @@ def process_data(df: pd.DataFrame, interval):
     # convert to datetime objects rather than strings
     df["author_timestamp"] = pd.to_datetime(df["author_timestamp"], utc=True)
     df["committer_timestamp"] = pd.to_datetime(df["committer_timestamp"], utc=True)
+    df.loc[df["author_timestamp"] == df["committer_timestamp"], "author_timestamp"] = None
 
     df_final = pd.DataFrame()
 
@@ -154,7 +155,6 @@ def process_data(df: pd.DataFrame, interval):
         df_hour = pd.DataFrame(hour, columns=["Hour"])
         df_final = df_hour.groupby(["Hour"])["Hour"].count()
     else:
-        print("here first")
         weekday = pd.concat([df["author_timestamp"].dt.day_name(), df["committer_timestamp"].dt.day_name()])
         df_weekday = pd.DataFrame(weekday, columns=["Weekday"])
         df_final = df_weekday.groupby(["Weekday"])["Weekday"].count()
@@ -186,15 +186,15 @@ def create_figure(df: pd.DataFrame, interval):
         font=dict(size=14),
     )"""
 
-    fig = px.bar(df, x=column, color_discrete_sequence=[color_seq[3]])
-    fig.update_traces(hovertemplate="%{y} Activity Count: %{x}<br>")
-    fig.update_yaxes(
+    fig = px.bar(df, y=column, color_discrete_sequence=[color_seq[3]])
+    fig.update_traces(hovertemplate="%{x} Activity Count: %{y}<br>")
+    fig.update_xaxes(
         categoryorder="array",
         categoryarray=order,
     )
     fig.update_layout(
-        yaxis_title=column,
-        xaxis_title="Activity Count",
+        yaxis_title="Activity Count",
+        xaxis_title=column,
         font=dict(size=14),
     )
 
--- a/pages/chaoss/visualizations/contrib_activity_cycle.py
+++ b/pages/chaoss/visualizations/contrib_activity_cycle.py
@@ -141,7 +141,7 @@ def process_data(df: pd.DataFrame, interval):
     # for this usecase we want the datetimes to be in their local values
     # tricking pandas to keep local values when UTC conversion is required for to_datetime
     df["author_timestamp"] = df["author_timestamp"].astype("str").str[:-6]
-    df["committer_timestamp"] = df["author_timestamp"].astype("str").str[:-6]
+    df["committer_timestamp"] = df["committer_timestamp"].astype("str").str[:-6]
 
     # convert to datetime objects rather than strings
     df["author_timestamp"] = pd.to_datetime(df["author_timestamp"], utc=True)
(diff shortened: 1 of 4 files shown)


pages/chaoss/visualizations/contrib_activity_cycle.py
```diff
@@ -0,0 +1,178 @@
...
+    logging.debug(f"{VIZ_ID} - END - {time.perf_counter() - start}")
+    return fig
+
+
+def process_data(df: pd.DataFrame, interval):
+
+    # for this usecase we want the datetimes to be in their local values
+    # tricking pandas to keep local values when UTC conversion is required for to_datetime
+    df["author_timestamp"] = df["author_timestamp"].astype("str").str[:-6]
+    df["committer_timestamp"] = df["author_timestamp"].astype("str").str[:-6]
```
JamesKunstle, line 144: I think that these are both going to be 'author_timestamp' rather than one being 'committer_timestamp'

pages/chaoss/visualizations/contrib_activity_cycle.py
```diff
@@ -0,0 +1,178 @@
...
+    return df_final
+
+
+def create_figure(df: pd.DataFrame, interval):
+
+    column = "Weekday"
+    if interval == "H":
+        column = "Hour"
+
+    fig = px.histogram(df, x=column, color_discrete_sequence=[color_seq[3]])
```
JamesKunstle, line 167: I believe that it's faster to pre-compute the count per category (per hour, per day) and then plot a bar chart of the counts than it is to use a histogram, so you could consider that optimization.

JamesKunstle (changes requested):
just a couple of comments on the process_data logic

pages/chaoss/visualizations/contrib_activity_cycle.py
JamesKunstle: since committer and author timestamps could be the same (and often are) you will double count if you concatenate them together. if they're not equal you should include both, if they are you should include only one. that means they're an author-committer.

cdolfi:
@JamesKunstle Let me know what you think of the changes and on the flipped graph view. Plotly generated that view out of box and once I saw it I liked it more but also can understand why to flip to normal. It was nice to see the bars less condensed and found it easier to compare

pages/chaoss/visualizations/contrib_activity_cycle.py
```diff
@@ -0,0 +1,201 @@
...
+    df["committer_timestamp"] = pd.to_datetime(df["committer_timestamp"], utc=True)
+
+    df_final = pd.DataFrame()
+
+    if interval == "H":
+        hour = pd.concat([df["author_timestamp"].dt.hour, df["committer_timestamp"].dt.hour])
+        df_hour = pd.DataFrame(hour, columns=["Hour"])
+        df_final = df_hour.groupby(["Hour"])["Hour"].count()
+    else:
+        print("here first")
```
JamesKunstle, line 157: should be removed

pages/chaoss/visualizations/contrib_activity_cycle.py
```diff
@@ -0,0 +1,201 @@
...
+    df["committer_timestamp"] = df["committer_timestamp"].astype("str").str[:-6]
+
+    # convert to datetime objects rather than strings
+    df["author_timestamp"] = pd.to_datetime(df["author_timestamp"], utc=True)
+    df["committer_timestamp"] = pd.to_datetime(df["committer_timestamp"], utc=True)
+
+    df_final = pd.DataFrame()
+
+    if interval == "H":
+        hour = pd.concat([df["author_timestamp"].dt.hour, df["committer_timestamp"].dt.hour])
```
JamesKunstle, line 153: This is still a double-count. If for a given commit the author is also the committer then the timestamp is counted twice, over-representing a time. If the committer isn't the author then both of the time events should be recorded because they're different.

Could screen out rows where author_timestamp == committer_timestamp and set those apart from where author_timestamp != committer_timestamp, and count the former only once each and the latter as you are now.

pages/chaoss/visualizations/contrib_activity_cycle.py
```diff
@@ -0,0 +1,201 @@
...
+        df_hour = pd.DataFrame(hour, columns=["Hour"])
+        df_final = df_hour.groupby(["Hour"])["Hour"].count()
+    else:
+        print("here first")
+        weekday = pd.concat([df["author_timestamp"].dt.day_name(), df["committer_timestamp"].dt.day_name()])
+        df_weekday = pd.DataFrame(weekday, columns=["Weekday"])
+        df_final = df_weekday.groupby(["Weekday"])["Weekday"].count()
+
+    # combine the weekday values for author and committer
+    """weekday = pd.concat([df["author_timestamp"].dt.day_name(), df["committer_timestamp"].dt.day_name()])
```
JamesKunstle, line 163: delete comment

pages/chaoss/visualizations/contrib_activity_cycle.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def create_figure(df: pd.DataFrame, interval):
+
+    column = "Weekday"
+    order = ["Saturday", "Friday", "Thursday", "Wednesday", "Tuesday", "Monday", "Sunday"]
+    if interval == "H":
+        column = "Hour"
+
+    """fig = px.histogram(df, x=column, color_discrete_sequence=[color_seq[3]])
```
JamesKunstle, line 178: delete comment

JamesKunstle (changes requested):

pages/chaoss/visualizations/contrib_activity_cycle.py
JamesKunstle: The above solution makes sense to me.

pages/chaoss/visualizations/contrib_activity_cycle.py
```diff
@@ -0,0 +1,201 @@
...
+
+    # for this usecase we want the datetimes to be in their local values
+    # tricking pandas to keep local values when UTC conversion is required for to_datetime
+    df["author_timestamp"] = df["author_timestamp"].astype("str").str[:-6]
+    df["committer_timestamp"] = df["committer_timestamp"].astype("str").str[:-6]
+
+    # convert to datetime objects rather than strings
+    df["author_timestamp"] = pd.to_datetime(df["author_timestamp"], utc=True)
+    df["committer_timestamp"] = pd.to_datetime(df["committer_timestamp"], utc=True)
+    df.loc[df["author_timestamp"] == df["committer_timestamp"], "author_timestamp"] = None
```
JamesKunstle, line 149: comment this for future readers + workshop attendees

pages/chaoss/visualizations/contrib_activity_cycle.py
```diff
@@ -0,0 +1,201 @@
...
+            categoryorder="array",
+            categoryarray=["Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"],
+        )
+    fig.update_layout(
+        yaxis_title="Activity Count",
+        font=dict(size=14),
+    )"""
+
+    fig = px.bar(df, y=column, color_discrete_sequence=[color_seq[3]])
+    fig.update_traces(hovertemplate="%{x} Activity Count: %{y}<br>")
```
JamesKunstle, line 190: I think the 'x' value could be templated as "Hour: {x}:00"; makes more sense than having the raw hour without context in the hover.

JamesKunstle (changes requested):

PR #311, opened 2023-04-26
augur app admin groups integration

JamesKunstle:
when a user has logged in and gotten the bearer token for their session, also query any groups from the ADMIN level- that is, those groups created by the account that the app is registered to via app_id and client_secret.


cdolfi (changes requested):
Going to go through and test this locally as well to see if it works as expected on my machine for the case when log in is use and when it isnt. I wanted to submit this back now to mention editing the augur_login.md for the new changes

cdolfi (changes requested):
After working with it locally these are the requested changes: 
- change "refresh" button to "Update Groups" and move the button to not be folded into the 
"More" button as it is not an intuitive user action (but best option we currently have available) 
- have "username_default" group available to access

JamesKunstle:
> After working with it locally these are the requested changes:
> 
> * change "refresh" button to "Update Groups" and move the button to not be folded into the
>   "More" button as it is not an intuitive user action (but best option we currently have available)
> * have "username_default" group available to access

The 'username_default' group isn't made unavailable- any group that doesn't include any repos is ignored when we load the groups. I'll change the button now.

cdolfi:
@JamesKunstle Group issues fixed in last update, Two requested changes now are: 
- editing the augur_login.md
- change "refresh" button to "Update Groups" and move the button to not be folded into the
"More" button as it is not an intuitive user action (but best option we currently have available)

JamesKunstle:
@cdolfi updated AUGUR_LOGIN.md and changed button name in pages/index/index_layout.py. I think moving the "update groups" button requires a redesign of the header of the page so that should be a separate styling PR.

JamesKunstle:
@cdolfi updated AUGUR_LOGIN.md and changed button name in pages/index/index_layout.py. I think moving the "update groups" button requires a redesign of the header of the page so that should be a separate styling PR.

pages/index/index_layout.py
```diff
@@ -20,7 +20,7 @@
                                 dbc.DropdownMenu(
                                     children=[
                                         dbc.DropdownMenuItem(
-                                            "Refresh",
+                                            "Refresh Groups",
```
cdolfi, line 23: Move refresh groups to its own tab (example "chaoss" on the other side) for user can see it without clicking in

cdolfi (changes requested):

pages/index/index_layout.py
JamesKunstle: as in above comment, should be a redesign PR, not this one

cdolfi: @JamesKunstle only necessary change to do this is remove the refresh groups as a dropdown menu item and add 
dbc.NavItem(
                                dbc.NavLink(
                                        "Refresh Groups",
                                        id="refresh-button",
                                    ),
                            ),

after(or before) the drop downmenu

cdolfi: @JamesKunstle I dont feel like its necessary to do a whole different pr for such a minor change. Let me know what you think after seeing the solution

JamesKunstle: @cdolfi I'm happy to do that but do we want to keep the dropdown if the only thing in there is the logout button? that's why I'm suggesting a redesign PR

cdolfi: @JamesKunstle My first instinct is to take it out completely but for this PR Im fine with just the refresh button being on the outside. I just think it is important to have it visible before going live with the log in as I think most/all users wont intuitively look for it and think the app is broken. Code to remove completely:
```
dbc.NavItem(
      dbc.NavLink(
              "Refresh Groups",
              id="refresh-button",
          ),
  ),
dbc.NavItem(
    dbc.NavLink(
            "Log out",
            id="logout-button",
        ),
), 
```

JamesKunstle: okay done

cdolfi (approved):
lgtm!

PR #313, opened 2023-04-29
alternate ports experiment for multiple instances on a compute

sgoggins:
Opening PR solely to help facilitate the parameterization of ports by exposing what I thought would work to eliminate this error when changing redis ports and eightknot ports: 
```
flower_1           |     raise ConnectionError(str(exc)) from exc
flower_1           | kombu.exceptions.OperationalError: Error 111 connecting to cache:6388. Connection refused.
callback-worker_1  | [2023-04-29 18:51:54,827: ERROR/MainProcess] consumer: Cannot connect to redis://cache:6388//: Error 111 connecting to cache:6388. Connection refused..
callback-worker_1  | Trying again in 6.00 seconds... (3/100)
```

### DIFF
--- a/cache_manager/cache_manager.py
+++ b/cache_manager/cache_manager.py
@@ -47,7 +47,7 @@ class CacheManager:
         self._redis = StrictRedis(
             # openshift will reconcile the 'redis' naming via the dns
             host=os.getenv("REDIS_SERVICE_HOST", "localhost"),
-            port=os.getenv("REDIS_SERVICE_PORT", "6379"),
+            port=os.getenv("REDIS_SERVICE_PORT", "6388"),
             password=os.getenv("REDIS_PASSWORD", ""),
             decode_responses=decode_value,
         )
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,7 @@ services:
       - ./env.list
     environment:
       REDIS_SERVICE_HOST: cache
-      REDIS_SERVICE_PORT: 6388
+      REDIS_SERVICE_PORT: 6380
     restart: always
 
   callback-worker:
@@ -26,7 +26,7 @@ services:
       - ./env.list
     environment:
       REDIS_SERVICE_HOST: cache
-      REDIS_SERVICE_PORT: 6388
+      REDIS_SERVICE_PORT: 6380
     restart: always
 
   query-worker:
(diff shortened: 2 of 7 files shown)


JamesKunstle:
@sgoggins I understand the request- I'll look over what needs to be done. Does the 8Knot instance need to be on a port other than 8050? I bet Flower needs to be on something other than 5555.

sgoggins:
Its actually just redis that doesn't appear to be working: 
```
flower_1           |     raise ConnectionError(str(exc)) from exc
flower_1           | kombu.exceptions.OperationalError: Error 111 connecting to cache:6388. Connection refused.
```

JamesKunstle:
So for multiple instances on the same machine, you'd like the ports to be non-colliding. I can automate that with a script that'll feed parameters to the docker-compose script so that instead of running 'docker-compose up --build' command you run './8knot_up.sh' and then the defaults will be '8050, 5555, and 6379' but you can override the defaults.

sgoggins:
> So for multiple instances on the same machine, you'd like the ports to be non-colliding. I can automate that with a script that'll feed parameters to the docker-compose script so that instead of running 'docker-compose up --build' command you run './8knot_up.sh' and then the defaults will be '8050, 5555, and 6379' but you can override the defaults.

Would I override the defaults in the 8knot_up.sh script? Or somewhere else?  Let me know when I can experiment with this. :)

JamesKunstle:
currently investigating automatically assigning the next available port up from 8050, 6379, and 5555 respectively. i.e. some netstat | grep '8050' kind of thing to iteratively find the next available port. 

otherwise the startup script will work like ssh-keygen w.r.t how you can just press enter through the defaults or change them if needed.

JamesKunstle:
It appears to be no problem to have two apps running via docker-compose by using the syntax:
'docker compose -p <unique-name> up --build'

challenge is that ports are mapped at build-time, so the 'cache' container's running redis instance expects to be mapped 6379:6379, as does flower and the webserver.

once option would be to have a script that does something like:

```bash

echo "Web-server port (default 8050):"
read webserver-port

echo "Flower port (default 5555):"
read flower-port

echo "Cache port (default 6379):"
read cache-port

echo "Application name:"
read app-name

echo "WEBSERVER-PORT=$webserver-port" > .build-env.sh
echo "FLOWER-PORT=$flower-port" >> .build-env.sh
echo "CACHE-PORT=$cache-port" >> .build-env.sh

....
```

sgoggins:
Ok. Let me know when I can try it. Looking to release next Tuesday.

PR #318, opened 2023-05-04
Repo button

cdolfi:


db_manager/augur_manager.py
```diff
@@ -275,6 +275,7 @@ def multiselect_startup(self):
         df_repo_git_id = df_search_bar.copy()
         df_repo_git_id = df_repo_git_id[["repo_git", "repo_id"]]
         self.repo_git_to_repo_id = pd.Series(df_repo_git_id.repo_id.values, index=df_repo_git_id["repo_git"]).to_dict()
+        self.repo_id_to_repo_git = {value: key for (key, value) in self.repo_git_to_repo_id.items()}
```
JamesKunstle, line 278: Vectorize with:
self.repo_id_to_repo_git = pd.Series(df_repo_git_id.repo_git.values, index=df_repo_git_id["repo_id"]).to_dict()

I checked it on my branch and it worked.

pages/index/index_layout.py
```diff
@@ -152,6 +152,14 @@
                     is_open=False,
                     color="info",
                 ),
+                dbc.Alert(
+                    children="List of repos",
+                    id="repo-list-alert",
+                    dismissable=True,
+                    fade=True,
+                    is_open=False,
+                    color="light",
```
JamesKunstle, line 161: style={"overflow-y": "scroll", "max-height": "440px"}

Makes the window scrollable and limits total height if there are a lot of entries. Window will be small if there are few.

JamesKunstle (changes requested):
Added two suggestions that I think make sense. 

First vectorizes the translation of the repo_id's to repo_gits based on what was being done before- replacement required because iterating takes a longer time than a vectorized option.

Second, makes the list of repo_gits scrollable and have a maximum height but not a minimum- a list of 5k will be scollable and some defined height but 1 repo will be a single line.

Something to investigate is how to dismiss the alert elegantly if the user stops focussing it- it sits above the search bar options. 

works for the time being, though.

Also: squash the PR into one commit rather than two or three.

pages/index/index_layout.py
sgoggins: @JamesKunstle : I think @cdolfi made the changes you requested ... Looking to get this tested today If I can.

JamesKunstle: I don't see the requested changes on the branch. I can merge and make my own PR for them though

PR #341, opened 2023-05-19
push admin groups query to client startup

JamesKunstle:
app-level defaults such as the groups that the admin user creates should be available to the user regardless of whether they have logged in.

This change moves the initial query for these groups to occur when the user first starts the application. They are also re-queried if the user signs in and requests to 'Refresh Groups' or if they restart the client from scratch.

This is a stop-gap implementation of this feature- in the future, we'll have a more refined login flow.


cdolfi (approved):
lgtm

PR #346, opened 2023-05-22
pr over time bug fix

cdolfi:
fix for now but need to get rid of the ".0" at the end of each closed

### DIFF
--- a/pages/overview/visualizations/pr_over_time.py
+++ b/pages/overview/visualizations/pr_over_time.py
@@ -188,6 +188,8 @@ def process_data(df: pd.DataFrame, interval):
         df_created["Date"] = df_created["Date"].dt.strftime("%Y-01-01")
         df_closed_merged["Date"] = df_closed_merged["Date"].dt.strftime("%Y-01-01")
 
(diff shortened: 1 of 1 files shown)


sgoggins (approved):
LGTM

JamesKunstle:
@cdolfi wait hold on, walk me through what the issue was and how this solves it. what I'm seeing is that 'closed' events should be different from 'merged' events, otherwise 'closed' will include 'merged' events because those are also 'closed'. however, I'm not sure if the operation takes the set-wise operation of 'closed minus merged = closed-but-not-merged'

the code may require some comments, I think that would help a lot.

cdolfi:
@JamesKunstle So this uses every "closed" and "merged" timestamp. If a PR is merged, then it is also considered "closed" at the same time. If a PR is "closed" (and not merged) it does not have a merged time stamp only closed. So since every merged pr is also technically a closed one, if you subtract the number of merged prs from the number of closed ones you just get the distinct values of PRs that were closed but not merged.

JamesKunstle:
@cdolfi we'll come back to this later to comment and cleanup

Issue #308, opened 2023-04-20
BUG: Lines metrics are not working  


cdolfi:

JamesKunstle:
I'm pretty sure that this is because the # of concurrent connections to Augur, either from the client or the server, is limited. Therefore when all of the callbacks to fill in those metrics fire they try to execute at once and a few are denied.

The patch for this is to background the metrics tasks. They're currently non-backgrounded callbacks so when the DB rejects the connection they just exit as failed. 

Backgrounding them will require a tiny bit of cache logic but will be a good idea because then they will only ever have to fire once for each repo and be updated on cache restart. When the cache is hot the values per repo will be accessed and returned in constant time.

cdolfi:
@JamesKunstle That makes sense, if we cant/dont have time to find a solution in the next few weeks we should temporarily take the metrics off of the hosted instance for OSSNA, RH Summit, etc

JamesKunstle:
@cdolfi Yeah we can do that if need be.

JamesKunstle:
UPDATE: 

We want to cache the metrics per repo- e.g. 'total number of commits for each repo selected' and then aggregate those statistics after caching. 

Therefore, the queries have to be updated to get the stats per selected repo rather than as one-shot for all repos.

JamesKunstle:
GOAL: 

Try to use existing data from queries for metrics rather than using whole new queries.

JamesKunstle:
finished as a part of Welcome page addition

PR #348, opened 2023-05-25
De-register 'Home' page of metrics pending overhaul

JamesKunstle:
Was working on #308 but I think this is a better decision in the long-run.

After having conversations about this page 3+ times, I've come to the conclusion that these 'time-agnostic-metrics' say little to nothing valuable about the projects they're representing. 

This should motivate a conversation about what 'single-number' metrics actually say something valuable; maybe we can take inspiration from Devstats and say "these are the metrics from the last {time-range}" and customize the time-range.

The following rationale is at the top of pages/home/home.py but I'm copying it here for the sake of clarity:

---
Single-value metrics like those currently represented aren't very useful, and feedback has
largely been that this page should say nothing rather than mis-represent project behavior.
These metrics currently capture the all-time behavior of a group of projects and say *something* without
being informative.

For example: {# lines added in all commits for all time} says nothing about the 
kind of contributions those lines were for, whether documentation or code. They aren't
good proxies for churn, because they're not isolated to any period of time.

Another example: {Avg. lifespan of PRs that have been merged for all time} is heavily skewed
by events like rubber-stamp merges at the beginning of a project's life, or extremely large overhauls
that take a long time to merge. It would be more useful to describe the IQR over time, binning the 
lifespans and considering outliers independently.  

Page will not be registered with Dash for the time being; to be replaced with a Welcome page.
---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Other stuff
 dictionary.dic
 tmp
+.vscode
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/pages/home/home.py
+++ b/pages/home/home.py
@@ -1,3 +1,23 @@
+"""HOMEPAGE DOCSTRING
+
+Page is pending overhaul.
+
+Single-value metrics like those currently represented aren't very useful, and feedback has
+largely been that this page should say nothing rather than mis-represent project behavior.
+These metrics currently capture the all-time behavior of a group of projects and say *something* without
+being informative.
+
+For example: {# lines added in all commits for all time} says nothing about the
+kind of contributions those lines were for, whether documentation or code. They aren't
+good proxies for churn, because they're not isolated to any period of time.
+
+Another example: {Avg. lifespan of PRs that have been merged for all time} is heavily skewed
+by events like rubber-stamp merges at the beginning of a project's life, or extremely large overhauls
+that take a long time to merge. It would be more useful to describe the IQR over time, binning the
+lifespans and considering outliers independently.
+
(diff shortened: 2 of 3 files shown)


cdolfi:
@JamesKunstle  I understand where you are coming from here, but with us pushing out the v1.0 email and promoting it to OCTO managers I feel like this makes the first view point of the application look less put together, and not sure how I feel about the info page being first especially since it is not populated. Lets talk more today about a solution.

JamesKunstle:
@cdolfi I should have been more explicit- Info is only being promoted the 'index' for the moment, because an index page has to exist, prior to an implementation of a 'Welcome' page, which is a draft issue I'm working on but haven't submitted. 

The 'Welcome' page will be the new 'index' page, with description of the different pages that are available and a guide to using the login functionality.

I don't feel good about highlighting this metrics page in a GA after some of the conversations we've had with people.

I'm open to include it but to demote it to a non-index page, such that a polished 'Welcome' page replaces it and gives users a soft intro to the app.

JamesKunstle:
@cdolfi

cdolfi:
@JamesKunstle Not merging as the new welcome page solution has not been implemented just yet

cdolfi:
@JamesKunstle Is this pr still relevant with #354 ?

JamesKunstle:
@cdolfi yeah the changes in this PR should be made following the merge of the 'Welcome' page, not before. I'll close this PR and reopen it when the Welcome page PR is made, which will deregister the Home page and promote the Welcome page to index instead.

PR #350, opened 2023-05-25
Find logs from workers by changing logging level defaults

### LINKED ISSUE #252: logging from celery query workers is lost
both in docker compose in deployment

JamesKunstle:
Fixes #252 

Logs from worker results were being hidden because they were at
DEBUG level rather than INFO level. This change promotes logs
generated by app functionality from L1 to L3, which supercedes
logs of L1 and L2 generated by the application framework.

Unfortunately, L3 is called 'Warning' which is too grave of an
intention for this change. That being said, we'll adopt the convention
that L3 logs, for the time being, are standard-functionality logs
because the TRACE logging level, old L1, has been deprecated.

TRACE = L1 (deprecated)
DEBUG = L2 (now L1)
INFO = L3 (now L2, which we want to be using)
WARNING = L4 (now L3, level that we're currenting using)
...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Other stuff
 dictionary.dic
 tmp
+.vscode
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/app.py
+++ b/app.py
@@ -23,7 +23,7 @@ from dash import CeleryManager, Input, Output
 import worker_settings
 import os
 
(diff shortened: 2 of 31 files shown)


cdolfi:
@JamesKunstle Once merge conflicts are solved Ill merge

cdolfi (changes requested):
Solve merge conflicts and it will be good to go

cdolfi (approved):
lgtm

PR #351, opened 2023-05-25
ignores 'search' action if nothing in searchbar

JamesKunstle:
fixes bug where search funcion allows searchbar to be empty. now raises dash.exceptions.PreventUpdate instead.


JamesKunstle:
fixes #295

cdolfi (approved):
lgtm

PR #352, opened 2023-05-31
filled in definitions on the info page

JamesKunstle:
have done 'contribution', 'contributor', 'contributor recency', 'contributor consistency', and 'staleness'

which covers most of the definitions we need for someone to use the application.

want to discuss what else needs to be on this page, but I think that the 'welcome' page should also be part of consideration because it'll help describe the functionality of the plotly graphs, a topic that was originally on the 'info' page.


pages/info/info.py
```diff
@@ -7,49 +7,151 @@
...
-        """
-        )
+        html.H3("Definitions"),
+        html.Hr(),
+        html.B("Contributor"),
+        html.Blockquote(
+            [
+                html.P(
+                    """
+                    An individual who has shared some software artifact or resource with an open source project.
```
cdolfi, line 17: For these definitions (atleast priorly) the idea was to give a "metrics definition" of what exactly is being counted on a graph where we say "contributor" (contribution is an easier example as there is a list of actions and contributor is one who does said actions). Im open to having the more general definitions, but think it is important to have both in that case. A qualitative and qualitative section might be a good approach to that

pages/info/info.py
```diff
@@ -7,49 +7,151 @@
...
+                    [
+                        html.Li("Starring a project"),
+                        html.Li("Forking a project"),
+                        html.Li("Watching a project"),
+                    ]
+                ),
+            ]
+        ),
+        html.Hr(),
+        html.B("Contributor Recency"),
```
cdolfi, line 58: Not sure im following this one. Whats the inspiration for this? Not sure I see this as a definition

pages/info/info.py
```diff
@@ -7,49 +7,151 @@
...
+                    Some contributions have a 'lifespan' between when they are opened and when they are closed.
+                    The 'lifespan' that a time-sensitive contribution to a project has can be very informative. If issues or pull requests are
+                    open for a very long time, for instance, this can signal problems for a community. Only those who are
+                    very familiar with a community can know how 'old' is a negative thing, however, and whether the 'age' of these
+                    contributions is important to consider.
+                    """
+                ),
+                html.P(
+                    """
+                    It should be obvious that this is a close relative to contributor activity states defined above.
```
cdolfi, line 132: Stating something as "obvious" can be demening. Different wording here or good with taking it out

pages/info/info.py
```diff
@@ -7,49 +7,151 @@
...
-            __Contribution__
-                > Definition for item here.
-
-            __Contributor__
-                > Definition for item here.
-
-            __Active Contributor__
-                > Definition for item here.
-
-            __Drifting Contributor__
```
cdolfi, line 24: I see the mapping between the different definition wording used below and these. I think it is important to have formal definitions to the exact terms used/measured in the graphs. This might be a part of the qualitative vs. quantitative breakdown I describe below

cdolfi (changes requested):

pages/info/info.py
JamesKunstle: Okay definitely, that's more useful- I'll just list everything we count.

pages/info/info.py
JamesKunstle: Good note- we have three groupings of terms:[Fresh, Staling, Stale], [Active, Drifting, Gone], and [Fly-by, Repeat] and since they don't mix it makes sense for them to be defined together. 

"Contributor Recency" is a term I would have brought up to you if you hadn't mentioned it here- I don't immediately think of a better term for "how long has it been since someone has contributed" but lmk if you do.

pages/info/info.py
cdolfi: That would work for me. I would just want to make sure to include specific verbiage about the differences between a general definition and a "metrics" or count definition. Imo the most important part of these definitions is to specify what we are counting in our different visualizations

pages/info/info.py
cdolfi: I dont have a different term for "contributor recency," for me that seems implied but dont mind including the definition. I do think its important to give context or definitions for the [Fresh, Staling, Stale], [Active, Drifting, Gone], and [Fly-by, Repeat] terms as we actively use them in our graphs

cdolfi (changes requested):

pages/info/info.py
JamesKunstle: I'm not sure I 100% understand what you mean- could you give an example?

pages/info/info.py
JamesKunstle: My opinion is that the description below captures the definitions well, at least insofar as the three words that we're using [Active, Drifting, Away] have very little literal meaning. If you have a specific disagreement I'd love to know- I can change how something is said but I think the definition should be exclusively literal about the recency terms describing time ranges.

pages/info/info.py
JamesKunstle: I'm not sure I totally understand what you're saying. Could you give an example of what you'd like to see included as well, and where?

cdolfi:
@JamesKunstle bumping

JamesKunstle:
@cdolfi I asked a couple questions in above threads. What additional info would you be looking for in these definitions?

pages/info/info.py
cdolfi: Instead of the "examples" describing it as "Contributions Measured" (or some type of wording that specifies what is counted in 8Knot). Then for the list it being each one individually listed (Open pull request, Close pull request, merge pull request...) or could be similar to what you have but changing the order for legibility to something like "Pull Requests: open, review, comment, close, merge"

cdolfi: and then specifying somewhere that an individual is counted as a contributor when one of the measure actions is completed

cdolfi (changes requested):
Small requested changes and then we can merge. Closed the comment on the [Fresh, Staling, Stale], [Active, Drifting, Gone], and [Fly-by, Repeat] terms, we should note somewhere to get feedback from OPSO like folks that arent as DS/math heavy if the verbiage make sense to them.

cdolfi (approved):
lgtm

PR #373, opened 2023-07-05
change to all queries to remove any future dates

cdolfi:


JamesKunstle (approved):
This is a simple review. I think this change makes a lot of sense. It doesn't seem like there would be a reason to prefer the Pandas timestamp object over Datetime, so this is an easy merge.

PR #378, opened 2023-07-26
removes use of pconfig, purely environment

### LINKED ISSUE #253: remove pconfig, use environment instead
currently passing db credentials to celery workers- should instead just use the present environment variables

JamesKunstle:
fixes #253
previously, pconfig was an object passed to celery workers that contained database access credentials in lieu of sending a global-scope instantiated AugurManager object.

this is replaced by created a new AugurManager object in each Celery worker process new from the available environment variables.

This should be more secure because we're not passing access parameters, in plaintext, between processes, as worker parameters.

### DIFF
--- a/app.py
+++ b/app.py
@@ -15,6 +15,7 @@ from db_manager.augur_manager import AugurManager
 import dash
 import dash_bootstrap_components as dbc
 from dash_bootstrap_templates import load_figure_template
+from sqlalchemy.exc import SQLAlchemyError
 import sys
 import logging
 import plotly.io as plt_io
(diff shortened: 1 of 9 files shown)


JamesKunstle:
Also deletes a lot of useless environment variable handling code in app.py that should have been dropped a long time ago.

Future task is to create an object that can handle AugurManager exception raising in a single line, removing the boilerplate code shared between all of the celery processes.

cdolfi (approved):
lgtm, really like the simplification

JamesKunstle:
@cdolfi cool, merged.

Issue #383, opened 2023-07-28
Audit of all visualizations that use contributor query 


cdolfi:
The contributor query now includes the following: 
- Commits 
- Issues: open, close, comment 
- Pull Requests: open, close, merge, review, comment

Check each visualization for use case on if anything needs to change with new information

cdolfi:
Audit complete: no changes necessary, will keep this in mind as new visualizations are built

PR #384, opened 2023-07-28
Update query for new values

cdolfi:
This query cleans up the new information and groups together the review information for the time being. #383 will be done to assess the review information use cases, as well as the other new actions.


JamesKunstle (approved):
lgtm, very reasonable.

PR #387, opened 2023-08-01
Contributor by Action Visualization

cdolfi:


JamesKunstle:
@cdolfi Seems like an open-and-shut review (very straightforward), but could you please update the commit message to describe what's being changed, what the rationale is, and then sign the commit with '-s' flag?

I think 'git reset' pops the HEAD commit so you can re-commit with 'git commit -a -s' which will stage everything back and then auto-sign the commit, putting you into the commit title/message writing interface. 

https://git-scm.com/docs/git-reset

Note of which issue this attends to, if any, would be listed as 'Fixes #<issue_number>'

Once that's done I'll review the code itself.

cdolfi:
@JamesKunstle with the limited characters for commit messaging i just added the issue associated

JamesKunstle:
@cdolfi Having the issue associated is good background.

I see that the vscode UI limits you to 50 characters of commit message. I also know that you don't really like using the terminal for Git stuff.

That being understood on my part, I'll just note that the character limit only applies to the commit title. 
Typing "git commit -s -m "<some commit title>" -m "<some commit body>" lets you write as much commit message body as you want, and the commit will be signed with your name as well.

Likewise, if you'd rather not write the commit message using the flags, 'git commit -s' will open a nano or vim window that will have the  commit title line character-limited, but allow you to write a longer commit message below. The '-s' will instruct the commit to be signed with your email at the bottom.

It's not that we really look back at the commit log much so this mostly a non-issue, but for the sake of convention, someday in the future we should enforce that commits (1) are signed, probably with GPG keys, (2) are adequately titled, (3) have a message body that explains the contents, and references the issue that the commit fixes, if such a thing exists.

pages/chaoss/visualizations/contribs_by_action.py
```diff
@@ -0,0 +1,229 @@
...
+                                                {"label": "PR Review", "value": "PR Review"},
+                                                {"label": "Issue Opened", "value": "Issue Opened"},
+                                                {"label": "Issue Closed", "value": "Issue Closed"},
+                                                {"label": "Commit", "value": "Commit"},
+                                            ],
+                                            value="PR Open",
+                                            clearable=False,
+                                        ),
+                                        dbc.Alert(
+                                            children="Please select different action as this set has no contributions of this type'",
```
JamesKunstle, line 74: Maybe a slight rewording:
"No contributions of this type have been made. Please select a different contribution type."

pages/chaoss/visualizations/contribs_by_action.py
```diff
@@ -0,0 +1,229 @@
...
+                html.H3(
+                    "Contributors by Action Type",
+                    className="card-title",
+                    style={"textAlign": "center"},
+                ),
+                dbc.Popover(
+                    [
+                        dbc.PopoverHeader("Graph Info:"),
+                        dbc.PopoverBody(
+                            """Visualizes the number of contributors in the specified time bucket that\n
```
JamesKunstle, line 34: I think it's really easy to misunderstand this one and think it's the # of a particular kind of contribution per time range. 

Maybe adjusting to:
"Visualizes the number of contributors who have performed a specific action (have opened a PR, for example) within a specified time-window. This is different from counting the number of contributions (the number of PRs having been opened)- the focus is on the activity of distinct contributors."

The adjustment is minor but I think the clarification is necessary.

pages/chaoss/visualizations/contribs_by_action.py
```diff
@@ -0,0 +1,229 @@
...
+
+    return df
+
+
+def create_figure(df: pd.DataFrame, interval, action):
+    # time values for graph
+    x_r, x_name, hover, period = get_graph_time_values(interval)
+
+    # create plotly express histogram
+    fig = px.histogram(df, x="created_at", color_discrete_sequence=[color_seq[3]])
```
JamesKunstle, line 202: Using a histogram means that the entire dataframe is transported to the client whenever this visualization is rendered. The alternative is to pre-aggregate the columns per time bin ahead of time, and switch to a bar chart.

I don't know if it's necessary to do that here, we've just done it in the past.

pages/chaoss/visualizations/contribs_by_action.py
```diff
@@ -0,0 +1,229 @@
...
+
+    # creates bins with interval size and customizes the hover value for the bars
+    fig.update_traces(
+        xbins_size=interval,
+        hovertemplate=hover + "<br>Contributors: %{y}<br><extra></extra>",
```
JamesKunstle, line 207: I'd also like to see the hover include the type of contribution being considered.

JamesKunstle (changes requested):
Very nice addition, love it. 
Made a couple of comments, mostly suggestions of rewording for precision or spelling.
One comment about performance (histogram vs. bar chart), will go with your intuited preference.

pages/chaoss/visualizations/contribs_by_action.py
cdolfi: Agreed, I think this is something we should keep in mind. Sometimes we do histograms, sometimes we do bars in different applications. This comment actually made me look into those and remind myself the "why" of each bc its not been for arbitrary reasons once I looked into it/reminded myself. For now I think we should keep this as a histogram as it allows the hover to hold the full date information for quarters and half year in a way we cant directly get by using to period (insert a lot of pandas date range and plotly use of dates compatibility things here here lol).

cdolfi:
@JamesKunstle Changes made, let me know if you notice anything else that should be modified

pages/chaoss/visualizations/contribs_by_action.py
JamesKunstle: Sounds very reasonable, I agree with your decision fully.

JamesKunstle (approved):
Everything lgtm, awesome work.

PR #395, opened 2023-08-10
Project velocity

cdolfi:
This implements the project velocity visualization with minor fixes to the contributor query (and the impacted visualization from this change)

### DIFF
--- a/pages/chaoss/visualizations/project_velocity.py
+++ b/pages/chaoss/visualizations/project_velocity.py
@@ -17,11 +17,6 @@ import datetime as dt
 import math
 import numpy as np
 
-"""
-(5) CONTEXT OF GRAPH
-
-"""
-
 
 PAGE = "chaoss"
 VIZ_ID = "project-velocity"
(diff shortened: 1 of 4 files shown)


pages/chaoss/visualizations/project_velocity.py
```diff
@@ -0,0 +1,352 @@
...
+import io
+from cache_manager.cache_manager import CacheManager as cm
+from pages.utils.job_utils import nodata_graph
+import time
+import datetime as dt
+import math
+import numpy as np
+
+"""
+(5) CONTEXT OF GRAPH
```
JamesKunstle, line 21: remove

pages/chaoss/visualizations/project_velocity.py
```diff
@@ -0,0 +1,352 @@
...
+            [
+                html.H3(
+                    "Project Velocity",
+                    className="card-title",
+                    style={"textAlign": "center"},
+                ),
+                dbc.Popover(
+                    [
+                        dbc.PopoverHeader("Graph Info:"),
+                        dbc.PopoverBody("INSERT CONTEXT OF GRAPH HERE"),
```
JamesKunstle, line 41: Missed this

JamesKunstle (changes requested):
Tiny changes requested 😁 
Otherwise, we've already agreed on the implementation in Rappel, so once that's done I feel great about this PR.

pages/chaoss/visualizations/project_velocity.py
cdolfi: @JamesKunstle  Do you have any wording suggestions for this? Thinking about it a bit but would be great to have input as well since this graph is so complex

JamesKunstle: Thankfully, the graph itself is pretty descriptive. I think you could link the CHAOSS definition, and that'd be pretty solid. My expectation is that most people will get the vibe pretty quickly. 

I'd maybe also add a recommendation to strongly consider filtering on the last 6-12 months. Velocity is instantaneous, as should be the visualization- e.g. Kubernetes would have a crazy velocity if we take its entire history into account 😃

cdolfi: Hm I see what you are saying about the time range. One thing that I am worried about with that is that people will assume the base visualization is all time without actually checking vs them having to select the time range themselves for a window view. Most of the applications of project velocity I see are all time

JamesKunstle (approved):
lol i stg I already reviewed and approved this, I totally imagined that.

PR #404, opened 2023-08-16
name change from drive by to drive thru

cdolfi:
"drive by" has a bit of a negative/loaded connotation. Drive-thru gives the same meaning without the connotation


JamesKunstle (approved):
lgtm

Issue #393, opened 2023-08-09
Enhancement: Deploy compose with load balancer ahead of web server


JamesKunstle:
Currently, we can't boot more than one web server container to scale performance because that container binds to the :8050 address.

Instead, a load balancer should sit between the web server and the target socket so that we can dynamically bind other web servers to the load balancer in the future, and the LB stays bound to the desired port.

JamesKunstle:
Made this work in a toy application.
Small document describing method: https://medium.com/@vinodkrane/microservices-scaling-and-load-balancing-using-docker-compose-78bf8dc04da9

cdolfi:
closed by #408

Issue #389, opened 2023-08-07
Refactor compose and Openshift service networking to exclusively use DNS


JamesKunstle:
Before I understood namespace by-name DNS discovery of Openshift services I was using the local ephemeral IP of Redis that is injected into downstream pods to connect Celery workers to cache.

Should update this so that discovery is only every done via the DNS, allowing the IPs to be ephemeral w/o consequences.

JamesKunstle:
Sometimes, the openshift deployed 8Knot instance hangs- no work is able to be done. 

I think this is because the Redis container restarts, and because its cluster IP is ephemeral, it is reassigned a new IP. When this happens, the workers cannot discover the new Redis container because they aren't configured to find it via the DNS. If the container were connected to via a Service, the OCP DNS would handle the bookkeeping to keep everything synchronized.

JamesKunstle:
UPDATE:

DNS in docker-compose works great. In a Python program on the same compose network,
`cache = Redis(host="redis", port=6379)` is all that's required.

The deployed Redis instance is given a port at random by the operating system, so there's no collision.

JamesKunstle:
With file structure:
```
../compose-testing
├── Dockerfile
├── README.md
├── app.py
├── docker-compose.yaml
└── nginx.conf
```

docker-compose.yaml
```
services:

  flask-server:
    build: .
    ports:
      - 5001
    depends_on:
      - redis-cache

  reverse-proxy:
    image: nginx:latest
    volumes:
      - ./nginx.conf:/etc/nginx/nginx.conf:ro
    depends_on:
      - flask-server 
    ports:
      - "5001:5002"

  redis-cache:
    image: redis:latest
    ports:
      - 6379
```

Dockerfile
```
FROM python:alpine
WORKDIR /flaskapp
COPY ./app.py .
RUN pip3 install Flask redis
CMD python3 app.py
```

nginx.config
```
user  nginx;
events {
    worker_connections   1000;
}
http {
        server {
              listen 5002;
              location / {
                proxy_pass http://flask-server:5001;
              }
        }
}
```

app.py
```
from flask import Flask
import uuid
from redis import Redis

cache = Redis(host="redis-cache", port=6379)
app = Flask(__name__)
app_id = str(uuid.uuid1())

@app.route("/")
def index():
    return f"Flask Server {app_id}"

@app.route("/get")
def get():
    val = cache.get("key")
    return f"got {val}: {app_id}"

@app.route("/set/<value>")
def set(value):
[...]

PR #408, opened 2023-08-16
clean up docker files, use dns, add load balancer

JamesKunstle:
Refactors docker-compose file to only expose
the nginx load balancer and the optional flower
monitoring utility. All other services remain unexposed and are assigned ports by the OS to avoid semantic service collisions (Redis running on 6379 won't collide).

Also cleans up Dockerfiles and enforces DNS-only service referencing, rather than IP-based.


JamesKunstle:
Closes #393, #389

cdolfi (approved):
lgtm!

Issue #114, opened 2021-11-24
New Visualization: Bus Factor 


cdolfi:
**Please describe the background and context for this new visualization**
The Bus Factor is a compelling metric because it visualizes the question "how many contributors can we lose before a project stalls?" by hypothetically having these people get run over by a bus (more pleasantly, how many would have to win in a lottery and decide to move on).

The Bus Factor is the smallest number of people that make 50% of contributions.
**Describe the perspective you'd like the final visual to give**
https://chaoss.community/metric-bus-factor/

**Describe the acceptance criteria for the issue and visualization to be complete**
when the issue is taken on, the people working on this should edit to describe the steps necessary to complete this as a visualization in the explorer dashboard

christinaexyou:
@JamesKunstle please assign me to this issue.

JamesKunstle:
Some other interesting parameters:

If the definition of bus factor is "smallest number of people that make YY% of contributions." let the user set YY as a parameter.

Calculate bus factor over time using a time window of a parameterized width and step, e.g. "in six month windows, sliding by four months as a time, from <date> to <later date>."

Allow the user to screen out names by some regular expression from the bus factor, e.g. '-bot' names.

Allow the user to screen out outlier contributors (bots that aren't labelled 'bot' perhaps) all at once, but list the contributors who are green-lit (extremely active contributors who are outliers and not bots).

JamesKunstle:
Christina's initial analysis is merged into Rappel. Cali and Christina are working on Bus Factor integration into 8K.

JamesKunstle:
UPDATE: PR for notebook is done, now using this issue to track integration of analysis into application.

cdolfi:
closed with line and pie chart visualizations merged

PR #411, opened 2023-08-17
Add contrib importance over time line graph

christinaexyou:
This PR is part 2 of addressing issue #114 by making the following changes:

1. Creates a line graph depicting contributor importance over time
2. Updates queries/contributors_query.py to query for contributor alias emails and returns the dataframe with an email column included
3. Updates chaoss/chaoss.py


pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,443 @@
...
+                    is_open=False,
+                ),
+                dcc.Loading(
+                    dcc.Graph(id=f"{PAGE}-{VIZ_ID}"),
+                ),
+                dbc.Form(
+                    [   dbc.Row(
+                            [ 
+                                dbc.Label(
+                                    "Action Type:",
```
cdolfi, line 59: Change organization on inputs: 
row 1: Action type and "Contributor substring filter" (shorter name= more space) (potentially remove action type all together, see other comments)
row 2: threshold 
row 3: Window width and step size (look at staling graph for how to line them up)
row 4: Date range and about graph button

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,443 @@
...
+def create_figure(df: pd.DataFrame, action_type, step_size):
+    # create plotly express line graph
+    fig = px.line(df, x="period_from", y="cntrb_importance", custom_data=["period_to"], text="cntrb_importance")
+    
+    # format x-axis ticks to be in Month Year format
+    fig.update_xaxes(fixedrange=True, tickformat="%B %Y", tickangle=90, dtick=f"M{step_size}")
+
+    # define x-axis and y-axis titles and intialize first x-axis tick to start at the user-inputted start_date
+    start_date = min(df['period_from'])
+    fig.update_layout(
```
cdolfi, line 432: Change to quarterly ticks (every 3 months), common in business vernacular. Also use shorted month names to take up less space. First Time Contributions Per Quarter graph is a good example of this

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,443 @@
...
+        if remaining_slice:
+            break
+        
+    # create a dataframe with period to, period from, contributor importance, and cntrb_ids as columns
+    df = pd.DataFrame(list(zip(period_from, period_to, cis)), columns = ['period_from', 'period_to', 'cntrb_importance'])
+    return df
+ 
+def create_figure(df: pd.DataFrame, action_type, step_size):
+    # create plotly express line graph
+    fig = px.line(df, x="period_from", y="cntrb_importance", custom_data=["period_to"], text="cntrb_importance")
```
cdolfi, line 425: I actually think it might be a good idea where all of the actions are plotted at once instead of selecting one. Then each action will have a different line on the same graph and users can deselect from the legend if they dont want to see them in few. An example of a graph with multiple plots is Pull Requests Over Time. @JamesKunstle what are your thoughts on this?

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,443 @@
...
+    # define x-axis and y-axis titles and intialize first x-axis tick to start at the user-inputted start_date
+    start_date = min(df['period_from'])
+    fig.update_layout(
+        xaxis_title=f'Timeline (stepsize = {step_size} months)',
+        xaxis=dict(tick0=start_date),
+        yaxis_title='Contributor Importance' 
+                )
+    
+    # format hover template to display the start and end date of the window
+    hovertemplate="<br>".join(["period from: %{x: %m-%d}", "period to: %{customdata[0]: %m-%d}"])
```
cdolfi, line 439: Format hover like  First Time Contributions Per Quarter

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,443 @@
...
+                                        min_date_allowed=dt.date(2005, 1, 1),
+                                        max_date_allowed=dt.date.today(),
+                                        initial_visible_month=dt.date(dt.date.today().year, 1, 1),
+                                        clearable=True,
+                                    ),
+                                    width="auto",)
+                            ],
+                        ), 
+                        dbc.Row(
+                            [   dbc.Label(
```
cdolfi, line 130: Curious to hear @JamesKunstle thoughts on this as well but I think the "window width and step size" might be over complicating this a bit and might be confusing for users. It took me a while to be able to understand what was happening and I have been involved in this development process

cdolfi (commented):
Please ignore my inline comments for the time being. I want to keep them there to go back to pending the direction we choose to take.I think that we can greatly simplify this graph concept to make it easier to read. I feel like the width size and step size being different is a little confusing, especially for our audience. I have two proposed options that I think would work to get the intended analysis:

1) A bar chart in the style of the Contributor by action type graph. The range slider is not a object but a slider as part of the graph itself User inputs: 
-action type 
-date interval radio items
-Threshold 
- Substring filter 

2) A line chart with a line for each action type of a single graph. The range slider is not a object but a slider as part of the graph itself. User Inputs: 
-Threshold 
-"Step" size by days (step = window) (different name for step, trying to think of one) 
- Substring filter 

@JamesKunstle @christinaexyou let me know what yall think

christinaexyou:
@cdolfi @JamesKunstle I like option 2 (line graph) as an alternative - I agree that having both window width and step size is confusing. I also think that having a slider is more user-friendly compared to manually inputting a start and end date.

cdolfi:
@christinaexyou Agreed on the rangeslider. Lets see what  @JamesKunstle says, i like the line graph as well. Gives a lot of info in a single view

pages/chaoss/visualizations/contrib_importance_over_time.py
JamesKunstle: Yep I totally agree- it's also cool to see many lines at once haha.

pages/chaoss/visualizations/contrib_importance_over_time.py
JamesKunstle: Since those are a part of the analysis, I'd prefer to keep them and document them well, and have the defaults be sensible. Having a hard-coded default doesn't make sense, nor does removing this part of the analysis.

JamesKunstle (commented):
answered two questions

JamesKunstle:
@christinaexyou @cdolfi 
I think having the step-size and the window-size are important to the analysis, but that they could be hidden and have sensible defaults when not visible. 

I do prefer the look of the line graph per type- having all the graphs in one place is cool. It looks great and is more informative.

cdolfi:
@christinaexyou at this point I think it is good to move forward with making the graph in general, with maybe a default of window = step for the time being. @JamesKunstle Lets have a further discussion about the step and window size. There is probably a middle ground here. Personally if the window and step size are still included i would say at default they would be equal to each other. This will allow for the initial view to be interpretable even without the context and understanding of step vs. window. Then if people understand it they can change the input to their liking

pages/chaoss/visualizations/contrib_importance_over_time.py
christinaexyou: Since the x-axes values are cyclical according to the step size, I spaced them out such that there's a tick for every step_size * 2 months.

christinaexyou:
The following changes were made:
1. Updated the line graph to display every action type
2. Changed organization on inputs 
3. Updated x-axes ticks to display every step size * 2 months and abbreviated month names
4. Set the default window width and step size to be 6 months
5. Lined up window width and step size input boxes
6. Set hover to display date and number of contributions
7. Created an alert that produces a warning if the step size is greater than the window width

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,436 @@
...
+    A contributor is ranked by how large of a proportion of all work they did - 
+    if someone authored 350 of 1000 commits, for instance, they'll account for 35% 
+    of the commits made. The list of percent contributions is ranked in descending order. 
+    The ordered list is greedily processed such that the smallest list of contributors is 
+    found that accounts for threshold percent of the work done for a given contribution type.
+   
+    returns the contributor importance, cntrb_id's and number of outliers for the given contribution type
+    '''
+    # identify and count outliers which we define as 2 standard deviations or more above from the mean
+    num_outliers = len(df[(stats.zscore(df["Action"]) > 2)])
```
cdolfi, line 299: num outliers not used remove

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,436 @@
...
+    
+    # calculate total number of contributions
+    t_cntrbs = df["Action"].sum()
+
+    # initilize empty list of to store contibutor ids for those who make up the contributor importance
+    cntrb_ids = []
+    # initialize cumulative percent of contributions
+    cum_per = 0
+
+    # add column name cntrbs to calculate the ratio of activity to the total amount of said activity
```
cdolfi, line 309: I dont really understand this step

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,436 @@
...
+    df["cntrbs"] = df["Action"] / t_cntrbs
+    df = df.reset_index()
+    
+    # iterate through each row in df and create a running sum of percent contributions
```
cdolfi, line 313: I might be mis interpreting this but I dont know if this is actually getting bus factor. From reading this is seems like it is getting a possible number of contributors that meets the threashold but I dont see how it for sure gets the minimum number.

cdolfi: Either way I feel like this can be greatly simplified as well by df put into this function, summing by cntrb_id, and ordering from greatest to least. Then multiply t_cntrbs by the threshold (lets call this cntrb_thresh) and sum the column with the contribution counts until it is greater than or equal to cntrb_thresh

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,436 @@
...
+    # if the window width is greater than time spanned by start date and end date, raise ValueError
+    if window_width > delta_month:
+        raise ValueError('The window width cannot be greater than the time spanned by the start and end date.')
+    
+    # if the step size is greater than window width, raise ValueError
+    if step_size > window_width:
+        raise ValueError('The step size cannot be greater than the window width.')
+     
+    window_width=relativedelta(months=window_width)
+    step_size=relativedelta(months=step_size)
```
cdolfi, line 348: This all can be GREATLY simplified. Look at the process_data function for the active_drifting_contributor or staleness graphs on the overview page. Use pd.date_range to get the "start date"/"period from" for each step and then make a df (just like lines 209 to 216 in issue_staleness) with that as a column, then you can make another column for the end date_period to by just adding the window size to the start date. From there just use a .apply in a very similar fashion as those visualizations

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,436 @@
...
+        # exit loop if the remaining slice has been calculated
+        if remaining_slice:
+            break
+        
+    # create a dataframe with period to, period from, contributor importance, and cntrb_ids as columns
+    df = pd.DataFrame(list(zip(period_from, period_to, cis)), columns = ['period_from', 'period_to', 'cntrb_importance'])
+
+    return df
+
+def process_data(df, patterns, threshold, start_date, end_date, window_width, step_size):
```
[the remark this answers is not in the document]
cdolfi: This is also where the bot filter, and start and end date should be done

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,436 @@
...
+    # calculate contributor importance over time for each of the action types and return results in df
+    df_commit_result = calc_cntrb_importance_over_time(df[df["Action"].str.contains("Commit")], patterns, threshold, start_date, end_date, window_width, step_size)
```
cdolfi, line 402: This should all be one dataframe with each action being a column

cdolfi (changes requested):
This all can be greatly simplified, there is a lot of repeat work/unnecessary loops. Let me know if you want to work on this together or talk through any of the comments!

christinaexyou:
Changes made: 
1. Changed order of input elements to: threshold, window width/step size, filter, date picker range
2. Updated threshold to display percentages
3. Updated `About Graph` to add context for window width and step size
4. Changed `process_data` function to return df with columns as action types and values corresponding to the contributor importance of each action type
5. Added legend title

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,436 @@
...
+                html.H3(
+                    id=f"graph-title-{PAGE}-{VIZ_ID}",
+                    className="card-title",
+                    style={"textAlign": "center"},
+                ),
+                dbc.Popover(
+                    [
+                        dbc.PopoverHeader("Graph Info:"),
+                        dbc.PopoverBody("""
+                                        Visualizes how contributor importance evolves over a specified time period,\n
```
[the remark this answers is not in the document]
cdolfi: Put bus factor sentence first

pages/chaoss/visualizations/contrib_importance_over_time.py
cdolfi: turn x axis to be parallel to the graph, allows more space for the graph

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,416 @@
...
+                       marker=dict(color=color_seq[4])
+                       )
+                        
+        ], 
+    )
+    
+    # define x-axis and y-axis titles and intialize first x-axis tick to start at the user-inputted start_date
+    start_date = min(df_final['period_from'])
+    fig.update_layout(
+        xaxis_title=f'Timeline (stepsize = {step_size} months)',
```
cdolfi, line 400: take out the month value here or on the title, a little repetitive

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,416 @@
...
+                                    html_for=f"patterns-{PAGE}-{VIZ_ID}",
+                                    width="auto",), 
+                                dbc.Col(
+                                    [
+                                        dcc.Dropdown(
+                                            id = f"patterns-{PAGE}-{VIZ_ID}", 
+                                            options = [
+                                                {"value": "bot", "label": "bot"},
+                                            ], 
+                                            value = ['bot'],
```
cdolfi, line 143: @JamesKunstle thoughts on this being default to filter out the string bot? Maybe something like "potential-bot-filter" ? or completely blank?

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,416 @@
...
+    df_final = process_data(df, 
+                            patterns, 
+                            threshold, 
+                            window_width, 
+                            step_size, 
+                            start_date, 
+                            end_date)
+    
+    # drop rows with NaN values 
+    df_final = df_final.dropna(axis=0)
```
cdolfi, line 255: Why drop rows with NAN values? What condition creates rows with NAN values?

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,416 @@
...
+    if end_date is None:
+        end_date = df["created_at"].max()
+
+    if patterns:
+        # remove rows where Login column value contains the substring 'bot'
+        mask = df['login'].str.contains("|".join(patterns), na=False)
+        df = df[~mask]
+    
+    # convert integer threshold to a float decimal
+    threshold = threshold / 10
```
cdolfi, line 321: use a python type conversion function instead of brute force

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,416 @@
...
+    
+    # convert to datetime objects rather than strings
+    df["created_at"] = pd.to_datetime(df["created_at"], utc=True)
```
cdolfi, line 324: Move this line to 310 and then the next two lines are not necessary

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,416 @@
...
+    start_date = pd.to_datetime(start_date, utc=True)
+    end_date = pd.to_datetime(end_date, utc=True)
+
+    # order values chronologically by COLUMN_TO_SORT_BY date
+    df = df.sort_values(by='created_at', ascending=True)
```
cdolfi, line 329: line 324 and 329 should be the first two in process_date

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,416 @@
...
+    df_count_cntrbs = df_in_range.groupby(['Action', 'cntrb_id'])['cntrb_id'].count().to_frame()
+    df_count_cntrbs = df_count_cntrbs.rename(columns={'cntrb_id': 'count'}).reset_index()
+     
+    # pivot df such that the column names correspond to the different action types and the values are the number of contributions of each contributor
+    df_count_cntrbs = df_count_cntrbs.pivot(columns='Action', values='count')
+
+    # initilize empty df with columns that correspond to the different action types to store contributor importance 
+    df_cntrb_importance = pd.DataFrame(columns=df_count_cntrbs.columns)
+    
+    # create a dictionary to store contributor importance for each action type
```
cdolfi, line 299: If you want to follow line for line same logic as staling_issues, store as list instead of dictionary, have for loop be on the list of actions in a set order

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,416 @@
...
+    period_from = pd.date_range(start=start_date, end=end_date, freq=f'{step_size}m', inclusive="both")
+    # store the period_from dates in a df 
+    df_time_intervals = period_from.to_frame(index=False, name="period_from")
+    # calculate the end of each interval and store the values in a column named period_from
+    df_time_intervals['period_to'] = df_time_intervals.apply(lambda row: row.period_from + relativedelta(months = window_width), axis=1)
+    # initialize df to store cntrb importance for each action type for each window
+    df_cntrb_importance_over_time = pd.DataFrame(columns = df['Action'].unique())
+
+    # for each starting period, calculate the cntrb importance for each action type
+    for _, row in df_time_intervals.iterrows():
```
cdolfi, line 341: This can be changed to be done using .apply (much more efficient compared to iterrrows). The logic used in issue_staleness graph can be applied here. Some guidance to get you started (or we can work through this on a call): 
- df_time_intervals is now df_final 
- the function is something like 
`df_final["Commit"], df_final["PR Opened"], df_status["PR Closed"].... (* all other action types here) = zip(
        *df_status.apply(
            lambda row: cntrb_importance_over_time(df, row.period_from, row.period_to, window_width, threshold),
            axis=1,
        )
    )`

-

cdolfi (changes requested):
Getting super close! Now looking at the code I see it is possible to use .apply. It should be super minor switches to get it to work. Iterate is very inefficient compared to .apply and only be used when absolutely necessary

pages/chaoss/visualizations/contrib_importance_over_time.py
christinaexyou: There are alot of NaN values for contributor importance observed during the start of the project which makes sense. However, for commits, sometimes contributors aren't identified i.e have None for their cntrb_id so we can't include their data in contributor importance. Should I drop rows with cntrb_id = None in process_data instead of dropping rows in df_final ?

pages/chaoss/visualizations/contrib_importance_over_time.py
christinaexyou: I don't think I can use type conversion for division. The threshold is selected by the `Slider` object in Dash which only accepts integers as values eg) 10, 20, 30, etc. So I have to divide it by 10 to get the decimal representation of a percentage eg) 0.10, 0.20, 0.30, etc. I'll clarify my comment

cdolfi: Sounds good, that makes a lot of sense with the additional context

pages/chaoss/visualizations/contrib_importance_over_time.py
cdolfi: Ill be curious to hear what @JamesKunstle says, I think it should be done at the beginning of calc_cntrb_importance . The reason for my thought of this is for the cntrb_importance is not skewed by all the contributors with NAN as an ID getting considered as 1 contributor. Luckily this should only impact commits

christinaexyou:
Changes made:

1. Replaced for loop in `process_data` with .apply
2. Tweaked `calc_cntrb_importance` and `cntrb_importance_over_time` so that they work in conjuction with .apply
3. Re-ordered and deleted redundant code for processing time series data in `process_data`
4. Edited the `Graph Info` description
5. Tweaked plot layout

Next steps:
- Need feedback on how to handle `cntrb_ids == None` for commits. Currently, I'm keeping them in the df which is why there are gaps in my graph for commits.

pages/chaoss/visualizations/contrib_importance_over_time.py
cdolfi: ^ bumping as still relevant

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,423 @@
...
+    # hover template styling 
+    fig.update_traces(textposition="top right", 
+                      hovertemplate= "Date: %{x}" + "<br>Contributor Importance: %{y}<br><extra></extra>",
+                      )
+    
+    # layout syling
+    fig.update_layout(
+        xaxis_title=f'Timeline (stepsize = {step_size})',
+        xaxis=dict(tick0=start_date),
+        yaxis_title='Contributor Importance',
```
cdolfi, line 417: Repeating as I wasnt able to comment on the last one and including message sent via chat: "turn x axis to be parallel to the graph, allows more space for the graph" 

So since the x axis values are at an angle currently they take up more space. Since this graph has so many inputs its size is already a lot smaller than most graphs so anything space saving is good. Honestly since it is already just ending up being "Oct 20xx" ticks with just the year makes sense. Personally I dont care too much as long as they can all fit without being rotated.

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,423 @@
...
+
+    # calculate the threshold amount of contributions 
+    thresh_cntrbs = df[action_type].sum() * threshold
+    
+    # initilize running sum of contributors who make up contributor importance
+    cntrb_importance = 0
+
+    # initialize running sum of contributions
+    running_sum = 0
+
```
cdolfi, line 279: The more I think about it here is where the nan columns should be dropped. Then the cntrb importance is not skewed

cdolfi (changes requested):
Everything is super close! I resolved all prior review comments so only the ones that still need to be addressed are still visible. Once the final review comments are handled and @JamesKunstle reviews and addresses the comments to him this is good to go!

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,423 @@
...
+    # threshold is an integer value eg. 10, 20,..., 90 since dcc.Slider only accepts integers as values
+    # divide by 10 to convert it to a decimal representation of a percentage eg. 0.10, 0.20,..., 0.90
+    threshold = threshold / 10
+    
+    # create bins with a size equivalent to the the step size starting from the start date up to the end date
+    period_from = pd.date_range(start=start_date, end=end_date, freq=f'{step_size}m', inclusive="both")
+    # store the period_from dates in a df 
+    df_final = period_from.to_frame(index=False, name="period_from")
+    # calculate the end of each interval and store the values in a column named period_from
+    df_final['period_to'] = df_final.apply(lambda row: row.period_from + relativedelta(months = window_width), axis=1)
```
cdolfi, line 339: use pd.date_offset instead

https://stackoverflow.com/questions/46741423/add-months-to-a-date-in-pandas
https://pandas.pydata.org/docs/reference/api/pandas.tseries.offsets.DateOffset.html

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,423 @@
...
+
+    
+    # hover template styling 
+    fig.update_traces(textposition="top right", 
+                      hovertemplate= "Date: %{x}" + "<br>Contributor Importance: %{y}<br><extra></extra>",
+                      )
+    
+    # layout syling
+    fig.update_layout(
+        xaxis_title=f'Timeline (stepsize = {step_size})',
```
JamesKunstle, line 415: Right now this just renders as numerical:

"Timeline (stepsize =6)" so an annotation about the unit of step size would be nice.

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,423 @@
...
+    df_final = period_from.to_frame(index=False, name="period_from")
+    # calculate the end of each interval and store the values in a column named period_from
+    df_final['period_to'] = df_final.apply(lambda row: row.period_from + relativedelta(months = window_width), axis=1)
+
+    # dynamically calculate the contributor importance over time for each of the action times and store results in df_final
+    (df_final['Commit'],
+    df_final['Issue Opened'], 
+    df_final['PR Opened'],
+    df_final['PR Comment'], 
+    df_final['PR Review']) = zip(*df_final.apply(lambda row: cntrb_importance_over_time(df, row.period_from, row.period_to, window_width, threshold), axis=1))
```
JamesKunstle, line 346: Wow, that's a super compact piece of code, very nice.

JamesKunstle (changes requested):
I feel good about this analysis in general- the statistical premise is as sound as it was when we were reviewing it in Rappel. 

My requested changes are more communication-oriented. This is a nuanced topic that we should try to communicate very clearly to users.

JamesKunstle:
@christinaexyou @cdolfi 

On re-reading these PR's I want to make a comment: I'd like to find another term to use in place of "Importance."

Volume of contribution does not always imply productivity or impact, so I'd like to separate those concerns from this analysis. Fundamentally, what we're doing is tracking the smallest possible subgroup whose contributions account for some percentage of the total volume of contributions.

Here are some alternatives that I think are good, with an asterisk by my personal favorite.

- Plurality
- Preeminent
- Principal
- Prolific *
- Primary

I like Prolific and Prolificacy because it's defined purely in terms of quantity, not quality or impact:
https://dictionary.cambridge.org/us/dictionary/english/prolific

"Primary", "Principal", "Core" all have some overloaded meaning either in literal engineering hierarchy or in our other analysis.

What do ya'll think?

christinaexyou:
@JamesKunstle @cdolfi How does "Contributor Prolificacy Coefficient in {window_width} Month Windows" for the title sound ? The addition of the coefficient implies that it is a parameterized metric.

JamesKunstle:
> @JamesKunstle @cdolfi How does "Contributor Prolificacy Coefficient in {window_width} Month Windows" for the title sound ? The addition of the coefficient implies that it is a parameterized metric.

I love it, minus the word 'coefficient,' purely for aesthetic reasons. I think it can safely be omitted and the rest is very nice.

cdolfi:
@christinaexyou @JamesKunstle All sounds good to me!

pages/chaoss/visualizations/contrib_importance_over_time.py
cdolfi: @christinaexyou @JamesKunstle

christinaexyou:
@cdolfi @JamesKunstle
Changes made:

1. Added functionality user to create new keywords to filter out contributors by
2. Changed graph title 
3. Updated `Graph Info` description
4. Updated x-axis tick marks to be vertical
5. Dropped contributors with cntrb_id == None in `calc_cntrb_prolificacy_over_time` since the inputted df in `calc_cntrb_prolificacy` does not have cntrb_id as one of its columns. It's already processed to have the action types as its columns and the number of contributions per contributor as its values
6. Updated threshold to allow user to percentages that go up by 5% from 10% to 95%

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,478 @@
...
+                        
+        ], 
+    )
+    
+    # define x-axis and y-axis titles and intialize first x-axis tick to start at the user-inputted start_date
+    start_date = min(df_final['period_from'])
+
+    # update xaxes to display ticks
+    fig.update_xaxes(showgrid=True,
+                     ticklabelmode="period",
```
cdolfi, line 458: Still showing as slanted, personally would just do the ticks to show year or only show "Oct 20xx" every other year

cdolfi: from prior comments parallel to graph = horizontal

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,478 @@
...
+    # threshold is an integer value eg. 10, 15,..., 90 since dcc.Slider only accepts integers as values
+    # divide by 100 to convert it to a decimal representation of a percentage eg. 0.10, 0.15,..., 0.90
+    threshold = threshold / 100
+
+    # create bins with a size equivalent to the the step size starting from the start date up to the end date
+    period_from = pd.date_range(start=start_date, end=end_date, freq=f'{step_size}m', inclusive="both")
+    # store the period_from dates in a df 
+    df_final = period_from.to_frame(index=False, name="period_from")
+    # calculate the end of each interval and store the values in a column named period_from
+    df_final['period_to'] = df_final.apply(lambda row: row.period_from +  pd.DateOffset(months = window_width), axis=1)
```
cdolfi, line 395: dont need to use lamda: 
df_final['period_to']= df_final.period_from + pd.DateOffset(months=window_width)

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,478 @@
...
+
+    return cntrb_prolificacy
+
+def cntrb_prolificacy_over_time(df, period_from, period_to, window_width, threshold):
+    # subset df such that the rows correspond to the window of time defined by period from and period to
+    time_mask = (df['created_at'] >= period_from) & (df['created_at'] <= period_to)
+    df_in_range = df.loc[time_mask]
+    
+    # drop rows where the cntrib_id is None
+    cntrb_mask = (df_in_range['cntrb_id'] == None)
```
cdolfi, line 347: By doing num here it is making the thresh_cntrbs in calc_cntrb_prolificacy incorrect as it doesnt include the contributions from contributors that are nulls

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,478 @@
...
+                                        [dcc.Slider(
+                                            id=f"threshold-{PAGE}-{VIZ_ID}",  
+                                            min=10, 
+                                            max=95, 
+                                            value=50,   
+                                            marks={i: f'{i}%' for i in range(10, 100, 5)}
+                                        ),    
+                                    ], 
+                                    className="me-2",
+                                    width=12,
```
cdolfi, line 78: change size of slider to allow "threshhold" to be on the same line to save space. Graph is already very small from all of the necessary inputs

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,478 @@
...
+@callback(
+    Output(f"patterns-{PAGE}-{VIZ_ID}", "options"),
+    Output(f"keyword-{PAGE}-{VIZ_ID}", "data"),
+    Input(f"patterns-{PAGE}-{VIZ_ID}", "search_value"),
+    State(f"patterns-{PAGE}-{VIZ_ID}", "options"),
+    State(f"keyword-{PAGE}-{VIZ_ID}", "data"),
+    prevent_initial_call=True
+)
+
+def dropdown_menu(key, options, previous_key):
```
cdolfi, line 221: leaving to @JamesKunstle expertise as he is much more experiences with dropdowns

cdolfi (changes requested):
Only a few small things from me and then @JamesKunstle review. Getting super close!

pages/chaoss/visualizations/contrib_importance_over_time.py
JamesKunstle: If this works then I'm not opposed to using it, though the dash_mantine_components 'MultiSelect' component implements a 'createable' parameter, which might solve the problem for free.
https://www.dash-mantine-components.com/components/multiselect

pages/chaoss/visualizations/contrib_importance_over_time.py
JamesKunstle: ++

JamesKunstle (changes requested):
made a note about dmc.MultiSelect but broadly I'm very happy with this analysis and the Info+Title, I think it's suitable to be finished after Cali's recommendations have been addressed!

pages/chaoss/visualizations/contrib_importance_over_time.py
cdolfi: I really like this solution, would save a lot of code. FYI if you are looking for the "clearable" option its on the component props tab not documentation

cdolfi: Code for this can be found on dev in cntrb pie chart!

christinaexyou:
Changes made:
1. Replaced dcc.Dropdown with dmc.MultiSelect
2. Changed width of threshold Slider 
3. Took pd.DateOffset out of lambda function
4. Rotated x-axis ticks to be horizontal and only displayed for every other year
5. Moved filtering out `cntrb_id == None`

pages/chaoss/visualizations/contrib_importance_over_time.py
```diff
@@ -0,0 +1,429 @@
...
+                            [   dbc.Label("Filter Out Contributors with Keyword(s) in Login:",
+                                    html_for=f"patterns-{PAGE}-{VIZ_ID}",
+                                    width="auto",), 
+                                dbc.Col(
+                                    [
+                                        dmc.MultiSelect(
+                                            id=f"patterns-{PAGE}-{VIZ_ID}",
+                                            placeholder="Bot filter values",
+                                            data=[
+                                                {"value": "potential-bot-filter", "label": "potential-bot-filter"},
```
cdolfi, line 144: take out potential-bot-filter not necessary anymore

cdolfi (changes requested):
One minor switch and to stay consistent to other graphs (and legibility) have the order of functions after the graph call back be: 

create_contrib_prolificacy_over_time_graph
process_data 
create_figure
cntrb_prolificacy_over_time
calc_cntrb_prolificacy

Once those are in and @JamesKunstle reviews the info this will be good to merge!

JamesKunstle (approved):
very nicely done- lgtm with the changes Cali requested.

cdolfi (approved):

PR #420, opened 2023-08-24
Update docker & docker compose

codekow:
Cleanup the Dockerfile and docker-compose


docker-compose.yml
```diff
@@ -1,92 +1,73 @@
 services:
 
-  reverse-proxy:
```
JamesKunstle, line 3: I would like to keep this part of the compose. It's here so that one can scale more web-server workers at once with `docker compose up --scale web-server=4` or whatever. If someone wanted to deploy 8Knot on bare metal this would be more convenient than having to reconfigure the gunicorn command.

JamesKunstle (changes requested):
I'd like to preserve the reverse proxy in the compose, otherwise this is very appreciated!

docker-compose.yml
codekow: The reverse proxy is back

JamesKunstle (approved):

JamesKunstle:
@cdolfi 
Two development notes:
(1) Add some "REDIS_SERVER" value in your env.list, like "REDIS_SERVER=123" or something for development,
(2) Use 'localhost:8080' instead of 'localhost:8050' to access the local instance.

PR #439, opened 2023-09-11
update: ns locations

codekow:


codekow:
Associated with this PR https://gitlab.com/osci/community-cage-infra-ansible/-/merge_requests/609

PR #442, opened 2023-09-13
Contributor Issue assignee Graph

cdolfi:
Things to note during review: 

- opendatahub-io is a good view of this graph (or any repo(s) that consistently do assignments) 
- Ctrb_id on the hover is on the side. Currently thats the only way I can see how to do it with the cntb_id not being static (spent multiple hours yesterday trying to figure it out)


pages/overview/visualizations/cntrib_issue_assignment.py
```diff
@@ -0,0 +1,331 @@
...
+                ),
+                dbc.Popover(
+                    [
+                        dbc.PopoverHeader("Graph Info:"),
+                        dbc.PopoverBody(
+                            """
+                            Visualizes the delta number of issue assignments for each contributor \n
+                            that meets the assignment criteria. This can help understand the workload'\n
+                            distribution between each contributor or maintainer over time.
+                            """
```
JamesKunstle, line 38: Suggestion:

"Visualizes the number of issues assigned to distinct contributors over time."

I can see in the function "issue_assignment" that you calculate the delta, but I'm not sure if that's strictly relevant to the info section. My read of this visualization is "For each contributor that has at least one issue assigned to them, how many issues were assigned to them per time bin?"

I'm slightly apprehensive about suggesting an interpretation in the Info section. I think the interpretation provided is reasonable for most cases but I'd be more in favor of leaving it to the user to decide what the data means to them. Not a strong feeling though, if we decided to go that route I'd probably offer an insight like this on every visualization- imo, we're either impartial data visualizations providers or we aren't, and I don't have a value judgement either way.

pages/overview/visualizations/cntrib_issue_assignment.py
```diff
@@ -0,0 +1,331 @@
...
+                    dcc.Graph(id=f"{PAGE}-{VIZ_ID}"),
+                ),
+                dbc.Form(
+                    [
+                        dbc.Row(
+                            [
+                                dbc.Label(
+                                    "Total Assignments Required:",
+                                    html_for=f"assignments-required-{PAGE}-{VIZ_ID}",
+                                    width={"size": "auto"},
```
JamesKunstle, line 56: I'm interested in your rationale for the value of this option. I don't intuitively see why someone would want to scope out contributors by their historical assignment volume but I may be missing some insight.

My perspective would be that I'd rather capture all assignments than miss any, although this option does allow for that. Maybe the default could be '1' for that case, and the user can decide if they have a more specific use-case than that.

JamesKunstle (changes requested):
Some questions about interface + wording but otherwise looks good.

pages/overview/visualizations/cntrib_issue_assignment.py
cdolfi: I personally think the delta is important because if people look for specifically the number of assignments over the time interval its not going to equal what we have 

Also why distinct? I feel like that is implied/not sure where this could be interpreted as non-distinct 

On providing context: personally I dont think data science is ever impartial, choosing what visualizations we provide inherently chooses to give a specific view on a community and each decision we make builds on that. Ill take it out though because I dont want to trigger a need to go through and update each info graph. I would say though we should consider in the future if we should because even if we dont put those things we will be repeating them ourselves to users when they come to us to help interpret.

pages/overview/visualizations/cntrib_issue_assignment.py
cdolfi: I found from playing with the intervals that this was the sweet spot of actually being able to see the graph and it not getting too granular for most use cases. From talking to people about how they would use this graph it would be to understand their teams workload so filtering out the users that only have a few makes it where the graph and legends are not over saturated. Id like to currate to the request and usecases we see and people can edit from there/we can change it if we get other feedback.

cdolfi:
@JamesKunstle lmk what you think, ill wait to do any changes on these two until we come to a consensus

pages/overview/visualizations/cntrib_issue_assignment.py
JamesKunstle: Okay that use-case makes sense, I can see what you mean.

JamesKunstle:
@cdolfi Here's a unified response to your responses:

I've become aware of the subtlety of this visualization that I wasn't aware of last night, even looking at the processing steps: this shows **new assignment events per contributor per bin**, rather than **number of issues contributor is assigned to per bin**. Is that correct?

If that's the case, then I'd suggest a corresponding title like "Issue Assignment Events" or something - the subtlety has to be squashed because the analysis being done is more interesting than a naive count of something like "In march, Cali had 4 issues assigned to her, and in April those issues were still assigned, so the count is still 4."

I'd also suggest an Info along the lines of:

"Visualizes the number of times an issue was assigned to a contributor, per contributor, per time-bin." 

This would clarify that what's being counted isn't the number of issues assigned to each contributor but rather the direct action of an issue being assigned. To me, that isn't immediately obvious when looking at the data without further context."

It's a comms problem way more than an analysis problem, and as always, we're trying to pack all the context into like, 4 words and 20 words for title and info, respectively haha.

cdolfi:
@JamesKunstle So this visualization actually does show all issues that are currently assigned to them during that time interval. With the processing steps, it first takes out all the issues created after the time interval and closed before had, therefore only leaving the issues currently open during that time interval (and that has ever been assigned to them in the first place but.. ill get to this later in this comment).  Then it filters out the assignments or unassignments that occurred after the time interval.  Then counts the assignments and unassignments to get the delta even if they were assigned before that time interval. 

This comment though did point out to me that I have not taken into account the issues that have never been assigned. This doesnt impact this visualization, but it greatly impacts the other one. Im going to change the query to work for both cases, so this has been extremely important logic for me revisit

cdolfi:
@JamesKunstle Updated the query. With the left join (which is necessary to get all the issue values), the query is much slower which makes sense. Once we are locked in on this query and the pr equivalent ill get a materialized view made of it

JamesKunstle:
@cdolfi I need to spend a bit more time with your PR now that I've read your latest comments- I don't think I intuited the nuance of the data initially. Will have all PR reviews for these four PRs in by end of my work day (will certainly be available at the start of your workday tomorrow.)

JamesKunstle (changes requested):
generally just recommending more documentation and some revision to the visualization 'info' section. otherwise, good to go imo.

pages/overview/visualizations/cntrib_issue_assignment.py
```diff
@@ -0,0 +1,335 @@
...
+    df_unassign = df_unassign[df_unassign["assign_date"] <= end_date]
+
+    # get all issue assignments
+    df_assigned = df_in_range[df_in_range["assignment_action"] == "assigned"]
+
+    # drop rows that have been assigned more recent than the end date
+    df_assigned = df_assigned[df_assigned["assign_date"] <= end_date]
+
+    # return the different of assignments and unassignments
+    return df_assigned.shape[0] - df_unassign.shape[0]
```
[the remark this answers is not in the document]
cdolfi: yes

cdolfi:
@JamesKunstle For some reason it is not letting me comment directly on your last one so Ill put it here. The count will never be negative because it is counting all assignments and un assignments up until the end date. 

Lines 314-320 take out the issue assignments on issues that where opened after the end date, and the issues closed before the start date. This leaves the df_in_range to only be the issue assignments on issues that are open during the time interval. 

- lines 258-261: Takes all issue events that are "unassigned" and open during the time interval. Then cuts off any unassignments that have happened after the end date. Then NOTE: it does not cut off the assignments that happened before the time interval 

- lines 258-261: Takes all issue events that are "unassigned" and open during the time interval. Then cuts off any unassignments that have happened after the end date. NOTE: it does not cut off the assignments that happened before the time interval

 It can be though of as a counter. For each assignment there is one more assigned to the contributor, and then for each un-assignment there is one less, so subtracting those two values gives you the number of assignments in that time interval. And this is only including assignments to that specific contributor (line 311).

pages/overview/visualizations/cntrib_issue_assignment.py
```diff
@@ -0,0 +1,335 @@
...
+    df_unassign = df_in_range[df_in_range["assignment_action"] == "unassigned"]
+
+    # drop rows that have been unassigned more recent than the end date
+    df_unassign = df_unassign[df_unassign["assign_date"] <= end_date]
+
+    # get all issue assignments
+    df_assigned = df_in_range[df_in_range["assignment_action"] == "assigned"]
+
+    # drop rows that have been assigned more recent than the end date
+    df_assigned = df_assigned[df_assigned["assign_date"] <= end_date]
```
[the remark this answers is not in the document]
cdolfi: This does not actually do the same thing but think I understand what you were trying to do

pages/overview/visualizations/cntrib_issue_assignment.py
```diff
@@ -0,0 +1,335 @@
...
+    df = df[df["assignee"] == contrib]
+
+    # drop rows that are more recent than the end date
+    df_created = df[df["created"] <= end_date]
+
+    # drop rows that have been closed before start date
+    df_in_range = df_created[df_created["closed"] > start_date]
+
+    # include rows that have a null closed value
+    df_in_range = pd.concat([df_in_range, df_created[df_created.closed.isnull()]])
```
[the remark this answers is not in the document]
cdolfi: applied idea

cdolfi:
@JamesKunstle ready for review based on comments

JamesKunstle:
> @JamesKunstle For some reason it is not letting me comment directly on your last one so Ill put it here. The count will never be negative because it is counting all assignments and un assignments up until the end date.
> 
> Lines 314-320 take out the issue assignments on issues that where opened after the end date, and the issues closed before the start date. This leaves the df_in_range to only be the issue assignments on issues that are open during the time interval.
> 
> * lines 258-261: Takes all issue events that are "unassigned" and open during the time interval. Then cuts off any unassignments that have happened after the end date. Then NOTE: it does not cut off the assignments that happened before the time interval
> * lines 258-261: Takes all issue events that are "unassigned" and open during the time interval. Then cuts off any unassignments that have happened after the end date. NOTE: it does not cut off the assignments that happened before the time interval
> 
> It can be though of as a counter. For each assignment there is one more assigned to the contributor, and then for each un-assignment there is one less, so subtracting those two values gives you the number of assignments in that time interval. And this is only including assignments to that specific contributor (line 311).
@cdolfi 

What happens if an issue, which is open across two time-windows, is assigned to a contributor in the first window and unassigned in the next window? 

e.g. For January, issue 1 is opened and assigned to contrib 'A', then in Febuary it's unassigned to 'A' but isn't assigned to anyone else. It also isn't closed in Feb. Wouldn't this lead to a negative count for contributor 'A'? They'd have a single 'unassignment' event for Febuary.

Admittedly, this is a corner case, and perhaps I'm missing something obvious in the data or analysis, but it seems like this is possible.

-- 
Once this nuance is solved, I want to turn attention back to the Info section, but I want to understand the above situation from your perspective first. I fear that I might come across as too picky with the 'Info' section on some visualizations, but that's only because I'm frequently frustrated with other people's descriptions in similar settings and want our to be very clearly better.

cdolfi:
> > @JamesKunstle For some reason it is not letting me comment directly on your last one so Ill put it here. The count will never be negative because it is counting all assignments and un assignments up until the end date.
> > Lines 314-320 take out the issue assignments on issues that where opened after the end date, and the issues closed before the start date. This leaves the df_in_range to only be the issue assignments on issues that are open during the time interval.
> > 
> > * lines 258-261: Takes all issue events that are "unassigned" and open during the time interval. Then cuts off any unassignments that have happened after the end date. Then NOTE: it does not cut off the assignments that happened before the time interval

^ see note 
> > * lines 258-261: Takes all issue events that are "unassigned" and open during the time interval. Then cuts off any unassignments that have happened after the end date. NOTE: it does not cut off the assignments that happened before the time interval
> > 

^ see note
> > It can be though of as a counter. For each assignment there is one more assigned to the contributor, and then for each un-assignment there is one less, so subtracting those two values gives you the number of assignments in that time interval. And this is only including assignments to that specific contributor (line 311).
> > @cdolfi
> 
> What happens if an issue, which is open across two time-windows, is assigned to a contributor in the first window and unassigned in the next window?
> 
> e.g. For January, issue 1 is opened and assigned to contrib 'A', then in Febuary it's unassigned to 'A' but isn't assigned to anyone else. It also isn't closed in Feb. Wouldn't this lead to a negative count for contributor 'A'? They'd have a single 'unassignment' event for Febuary.

No, see above. Let me know if you want to hop on a call if it would be easier for me to explain the logic verbally 

> 
> Admittedly, this is a corner case, and perhaps I'm missing something obvious in the data or analysis, but it seems like this is possible.
> 
> -- Once this nuance is solved, I want to turn attention back to the Info section, but I want to understand the above situation from your perspective first. I fear that I might come across as too picky with the 'Info' section on some visualizations, but that's only because I'm frequently frustrated with other people's descriptions in similar settings and want our to be very clearly better.

JamesKunstle:
@cdolfi I read your note and the code thoroughly- I'm okay hopping on a call later but please read my side carefully too, because I'm confused where we're not connecting here.

---
**Here's how the code reads in plain english:**

Line 330: "Consider only issue assignment events for a specified contributor"

Line 333: "Only consider issues that were created before the end-date"

Line 336: "Only consider issues that were closed after the start date, or those that haven't ever been closed"

Line 339->341: "Get rows in df_in_range where the event that occurred was 'unassignment' and the action took place before the end-date"

Line 344->346: "Get rows in df_in_range where the event that occurred was 'assignment' and the action took place before the end-date"

---

With the above reading, it's feasible that a specific contributor could be assigned an issue in time bucket 1, have a `df_assigned.shape[0] - df_unassign.shape[0]` of exactly 1 for that bucket, and then in bucket 2 which follows bucket 1, be unassigned exactly one issue (that same issue) and have a `df_assigned.shape[0] - df_unassign.shape[0]` of exactly -1. 

Within the same bucket it's obvious that an (assignment+unassignment) pair would contribute to a sum of 0, but I'm saying that it's possible for the value to be negative.

---

Another example: 

In Bin 1, a contributor is assigned 50 issues.(Assignments-Unassignments) is 50 for that bin. Those issues obviously exist within the time range, and are never closed.

In Bin 6, which is 5 bins later than Bin 1, that contributor has done no work, so those 50 issues are unassigned from them. (Assignments - Unassignments) is -50 for that bin.

---

IMO this is a feature, not a bug, because seeing negative assignments for some bin is interesting in itself.

JamesKunstle:
@cdolfi Okay take-two:

I understand now- the missing piece was not catching the detail "...does not cut off the assignments that happened before the time interval"

I see now that for each successive time-bin, unassignments are matching previously counted assignments, and new assignments without paired unassignments increase the count for that bin.

JamesKunstle (approved):
w/ new understanding, lgtm.

PR #444, opened 2023-09-13
Contributor pull request review assignment

cdolfi:
Things to note during review: 

- Ctrb_id on the hover is on the side. Currently thats the only way I can see how to do it with the cntb_id not being static (spent multiple hours yesterday trying to figure it out)


JamesKunstle:
@cdolfi My comments on #442 apply here nearly directly- same rationale, pretty willing to just merge this after we've solidified the Info section and the interface.

JamesKunstle (changes requested):
in other comment

JamesKunstle:
@cdolfi same comments now as for #442

cdolfi:
@JamesKunstle ready for review based on comments

JamesKunstle:
@cdolfi responded to both in #442, very nearly at consensus

JamesKunstle (approved):
lgtm w/ new understanding

PR #445, opened 2023-09-13
Issue assignments

cdolfi:
Things to note during review: 

- opendatahub-io is a good view of this graph (or any repo(s) that consistently do assignments) 

### DIFF
--- a/pages/overview/visualizations/issue_assignment.py
+++ b/pages/overview/visualizations/issue_assignment.py
@@ -23,7 +23,7 @@ gc_issue_assignment = dbc.Card(
         dbc.CardBody(
             [
                 html.H3(
-                    "Issue Assignment",
+                    "Issue Assignment Status Counts",
                     className="card-title",
                     style={"textAlign": "center"},
                 ),
@@ -239,6 +239,11 @@ def create_figure(df: pd.DataFrame, interval):
 
 
 def issue_assignment(df, start_date, end_date):
+    """
+    This function takes a start and a end date and determines how many
+    issues that are open during that time interval are assigned and
+    unassigned.
+    """
 
     # drop rows that are more recent than the end date
     df_created = df[df["created"] <= end_date]
@@ -250,13 +255,13 @@ def issue_assignment(df, start_date, end_date):
     df_in_range = pd.concat([df_in_range, df_created[df_created.closed.isnull()]])
 
     # get all issue unassignments
-    df_unassign = df_in_range[df_in_range["assign"] == "unassigned"]
+    df_unassign = df_in_range[df_in_range["assignment_action"] == "unassigned"]
 
     # drop rows that have been unassigned more recent than the end date
     df_unassign = df_unassign[df_unassign["assign_date"] <= end_date]
 
     # get all issue assignments
-    df_assigned = df_in_range[df_in_range["assign"] == "assigned"]
+    df_assigned = df_in_range[df_in_range["assignment_action"] == "assigned"]
 
     # drop rows that have been assigned more recent than the end date
     df_assigned = df_assigned[df_assigned["assign_date"] <= end_date]
--- a/pages/overview/visualizations/issue_assignment.py
+++ b/pages/overview/visualizations/issue_assignment.py
@@ -260,6 +260,9 @@ def issue_assignment(df, start_date, end_date):
     # drop rows that have been unassigned more recent than the end date
     df_unassign = df_unassign[df_unassign["assign_date"] <= end_date]
 
+    # include rows that have a null assignment value, therefore never assigned
+    df_unassign = pd.concat([df_unassign, df_in_range[df_in_range.assignment_action.isnull()]])
+
     # get all issue assignments
     df_assigned = df_in_range[df_in_range["assignment_action"] == "assigned"]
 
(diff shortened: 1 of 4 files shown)


pages/overview/visualizations/issue_assignment.py
```diff
@@ -0,0 +1,265 @@
...
+
+PAGE = "overview"
+VIZ_ID = "issue_assignment"
+
+gc_issue_assignment = dbc.Card(
+    [
+        dbc.CardBody(
+            [
+                html.H3(
+                    "Issue Assignment",
```
JamesKunstle, line 26: Maybe a tiny bit more detail here for a title?

"Issue Assignment Status Counts" or something

pages/overview/visualizations/issue_assignment.py
```diff
@@ -0,0 +1,265 @@
...
+    return fig
+
+
+def issue_assignment(df, start_date, end_date):
+
+    # drop rows that are more recent than the end date
+    df_created = df[df["created"] <= end_date]
+
+    # drop rows that have been closed before start date
+    df_in_range = df_created[df_created["closed"] > start_date]
```
JamesKunstle, line 247: Isn't this including rows where the closed date is after the start date because of the direction of the inequality?

pages/overview/visualizations/issue_assignment.py
```diff
@@ -0,0 +1,265 @@
...
+        xaxis_title="Time",
+        yaxis_title="Issues",
+        legend_title="Types",
+        font=dict(size=14),
+    )
+
+    return fig
+
+
+def issue_assignment(df, start_date, end_date):
```
JamesKunstle, line 241: docstring.

It's difficult to understand what's being counted in this function. Grouping operations and describing the data in comments is nice to have when reading code like this since we can't see the data without print statements.

queries/issue_assignee_query.py
```diff
@@ -0,0 +1,110 @@
...
+        return None
+
+    query_string = f"""
+                    SELECT
+                        i.issue_id,
+                        r.repo_id as id,
+                        i.created_at as created,
+                        i.closed_at as closed,
+                        ie.created_at AS assign_date,
+                        ie.action AS assign,
```
JamesKunstle, line 43: I'm guessing the action is 'assigned' or 'unassigned', so I think this variable name is somewhat confusing. Maybe "assignment_action" or something would be easier to understand in context in the processing logic.

JamesKunstle (changes requested):
couple of code comments.

pages/overview/visualizations/issue_assignment.py
cdolfi: Yes thats the purpose. Comment pretty much says exactly what you are saying but from the opposite perspective

pages/overview/visualizations/issue_assignment.py
cdolfi: So for me that title does not add any more context, feels a little repetitive. Open to different ideas though Ill think about it too, Only thing I have is "Issue Assignments over Time

pages/overview/visualizations/issue_assignment.py
JamesKunstle: The line of code is totally right, my b, fresh eyes.

pages/overview/visualizations/issue_assignment.py
```diff
@@ -0,0 +1,273 @@
...
+                    "Issue Assignment Status Counts",
+                    className="card-title",
+                    style={"textAlign": "center"},
+                ),
+                dbc.Popover(
+                    [
+                        dbc.PopoverHeader("Graph Info:"),
+                        dbc.PopoverBody(
+                            """
+                            Visualizes the number of assigned and unassigned issues in each \n
```
JamesKunstle, line 35: Reading the analysis, my takeaway is that this info section isn't 100% correct. We're not counting, at the end of the time window, how many issues are in the 'assigned' state, vs. the number in the 'unassigned' state. 

We're counting, over the duration of the time window, how many state changes occurred into both targets, coming from the shape of the DataFrame that lists all 'assignment' and 'unassignment' events that are within the date-range specified.

To be ultra-clear, for the same issue, multiple 'assign' and 'unassign' events can happen within the time range, at least from my understanding of the data. Since we're doing a left-outer join on the issues table, each issue will be represented multiple times, once for every matching event, or once with a null value for events if no events exist.

Therefore, the info section should reflect that we're really counting events:

"Visualizes the number of times issues were assigned and unassigned to contributors for each time-bin."

JamesKunstle (changes requested):
changing info section, otherwise lgtm

cdolfi:
@JamesKunstle Fixed code to make it where the values are how many issues are assigned or unassigned

pages/overview/visualizations/issue_assignment.py
```diff
@@ -0,0 +1,288 @@
...
+    # get all issue assignments and drop rows that have been assigned more recent than the end date
+    num_assigned_actions = df_in_range[
+        (df_in_range["assignment_action"] == "assigned") & (df_in_range["assign_date"] <= end_date)
+    ].shape[0]
+
+    # number of assigned issues during the time interval
+    num_issues_assigned = num_assigned_actions - num_unassigned_actions
+
+    # number of unassigned issues during the time interval
+    num_issues_unassigned = num_issues_open - num_issues_assigned
```
JamesKunstle, line 285: ah that clears things up to me.

JamesKunstle (approved):
v nice, lgtm

PR #467, opened 2023-09-25
Page reformat

cdolfi:
Overall I feel like the structure of this is very close but not tied down to anything. Looking for any feedback here 

Specific things that I would like feedback on: 

- Naming of "By Type" and "By Action" - I think the idea of the breakdown between these works but not sold on the naming. It would be logical for me or others to think type = contribution type. 
- Location of the by_type and by_action files. This is a solution I stumbled upon and liked but could see reasons for different solutions 
- If Contributor Pull Request Review Assignment and Issue should also be in the contributor by action page (whatever it ends up getting named) 
- descriptions for contributor page on welcome page 
- how to do the flex wrap on the welcome page. Might be best to have you do that on a separate PR


JamesKunstle (changes requested):
This is within a razor's margin of being ready to merge. Just need to agree on some of the language, as you @cdolfi state in your PR comment. 

I'm going to clone your branch, put myself in the shoes of a user, and make some suggestions based on that experience.

JamesKunstle:
Here's my two cents looking this over:

1. What if we rename 'Contributions' to 'Activity'? I'm imagining myself as the user, and to me, "Activity" makes the most sense for the general term for "things that are happening in the community," especially given that the visualizations are considering long-term trends.
2. I think that 'Affiliations' should be another tab under 'Contributors' since Affiliations can only ever be about contributors.
3. I'm having a hard time slicing 'By Type' vs. 'By Action' into meaningful subdivisions. One that sort of makes sense is "Behavior" vs. "Productivity", generally, separating the community statistics about the contributors (time when people are contributing, number of new contribs per window, engagement, repeat vs. fly-by) from productivity trends (Contributors by Action Type, prolificacy, first-time contribs, drive-by contribs). 
4. I'm also realizing that it'd be nice to have the page's context above the figures so that when we're clicking between drop-down items we know which page we're on w/o looking at the URL.

cdolfi:
@JamesKunstle 
1. Personally I feel like Activity is too general. The two points you make about "things that are happening in the community," especially given that the visualizations are considering long-term trends" apply to every visualization we have. Personally I feel like the "Contribution" title makes sense as its looking at counts and trends of contributions and seeing it next to contributors helps make the difference between the two obvious. 

2. I like having affiliations outside of it as many times we are asked about this topic separate from any other analysis and have a feeling we will get asked "where is it" if its under a drop down. If you have a strong feeling about this though im fine with putting it under affiliations 

3. The breakdown I see between the two are 
- By type: has to do with labels given to a contributor based on their behavior, so maybe behavior is the way to go
-by action: had to do with contributors based on the types of contribution types. Productivity doesnt really fit that for me. Since we have a replacement for the other "Contribution Type" makes sense for a title here 

Overall though I think we should choose something and get it up on main and we will be able to get feedback which will probably get us to a better place than just us two going back and forth

4. Lets open a separate issue / PR to handle this

JamesKunstle:
@cdolfi 
1. Ultimately I don't feel super strongly about this point- I do think that 'Contributions' and 'Contributors' as sections next to each other are slightly hard to distinguish in the UI on a 14" laptop, which is the impetus behind this revision suggestion. 'Activity' isn't a perfect title either. I'm fine w/ keeping 'Contributions' in the short run but I want to note my acknowledgement of the visual discernment challenge.
2. That's fine, I do think that categorically 'Affiliations' SHOULD be under contributors but you're right, that could easily become a rough user experience if people aren't looking thoroughly.
3. The challenge I have w/ the subtabs as they are now is that the visualizations themselves don't split into nice categories down a single line. As a user, I'm flipping between the two pages w/o really understanding why two pages exist in the first place- unless well explained, the two groupings seem to be arbitrary. 'Behavior' and 'Contribution Type' are sufficient in the short term IMO, but as you say, I's like to have some feedback on how people would actually slice those perspectives.

cdolfi:
@JamesKunstle Updates are in based on our discussion, let me know what you think. We are on the same page with what you said in your last comment. I opened issues #473 #472 to document each. Could you add a screenshot of what you are seeing on your laptop on #473?

JamesKunstle (approved):
lgtm!

Issue #470, opened 2023-09-26
Contributor importance line chart hover update 


cdolfi:
Update hover to include start and end date for the window

Issue #468, opened 2023-09-26
BUG: Contributor Importance pie chart and line chart not showing issue closed data 


cdolfi:
"Issue Closed" is not being shown in cases where issue closed data does exist

cdolfi:
closed by #479

PR #474, opened 2023-09-27
Add issues closed line graph and update hovertemplate

christinaexyou:
@cdolfi Resolves part of #468 and #470 by making the following changes:

* Adds line for issues closed
* Updates the hover template with `Date`, indicating the time window for which prolificacy is calculated


cdolfi (changes requested):
Relocate the visualization back to its location on dev currently and ill review. When opening a PR its good to remember to check to make sure you are not overwriting other changes from the current state of the branch and that the app runs before pushing!

PR #482, opened 2023-09-29
adds oauth public config to openshift definitions

JamesKunstle:
8Knot app needs these definition in order to connect to the Augur frontend authorization client for login flow.


cdolfi:
@JamesKunstle Does the readme need to be updated or is this an openshift only thing?

PR #484, opened 2023-10-02
adds redis instance for user session data to OCP

JamesKunstle:
'redis-users' service available in docker-compose definition, but explicit creation of another service for this purpose is required in the openshift resource definitions.

This commit includes these definitions and sets the environment variables required so that services can access the new cache instance.


cdolfi (approved):
lgtm!

PR #486, opened 2023-10-02
hover update

cdolfi:
After playing with this a bit, from what I have been able to figure out this is the only way at this point to get the legend information in the hover when using a histogram.


JamesKunstle (approved):
Okay, lgtm.

Issue #487, opened 2023-10-02
Materialized view vs. 8Knot query - repo_id


JamesKunstle:
I'm getting DB read failures on my branch, and I rooted out the problem:

The materialized views for `explorer_pr_assignments.sql` and `explorer_issue_assignments.sql` have the line `*.repo_id AS id,` rather than `*.repo_id as repo_id` or otherwise, so the queries in 8Knot are failing, expecting to find a `repo_id` column.

Unclear whether the 8Knot queries need to adopt the 'id' column name or the materialized views should have the column be 'repo_id.'

JamesKunstle:
Fix not necessary- materialized view conforms to other queries in app, just one step later than normal.

PR #488, opened 2023-10-02
fix issue_assignee_query and pr_assignee_query

JamesKunstle:


cdolfi:
@JamesKunstle is this still relevent / necessary?

cdolfi (approved):
still a little confused as to why this doesnt work on your branch in current state as it does on mine (even after rebuild) but they are equivalent so lgtm

JamesKunstle:
> https://github.com/oss-aspen/8Knot/issues/487 confused bc of the closing and comment of this

#487 addresses the problem but misidentifies the cause as being from the materialized view. 

This PR just fixes up the application query to be correct SQL, keeping the materialized view as-is.

JamesKunstle:
> still a little confused as to why this doesnt work on your branch in current state as it does on mine (even after rebuild) but they are equivalent so lgtm

Did you delete the existing Redis container?

`docker compose down; docker compose up --build`

The query wasn't fully correct SQL so it shouldn't work anywhere- it doesn't work in dbeaver. I'm guessing the cache container was still warm despite rebuild, which would happen if you didn't `down` the compose.

cdolfi:
sounds good im good to merge

PR #491, opened 2023-10-04
minor bug fix

cdolfi:
Stumbled upon this bug with the app opening on a non fully collected on repo. One of the visualizations were missing the extra input which was causing the catch for no data to fail


JamesKunstle (approved):
dope nice catch

PR #492, opened 2023-10-04
set the searchbar default from the environment

JamesKunstle:
allows deployer to set an environment variable for the label of the repo/org they'd like the searchbar to default to on startup.

If the variable not set, default is the first option available in the options list.

### DIFF
--- a/8Knot/db_manager/augur_manager.py
+++ b/8Knot/db_manager/augur_manager.py
@@ -55,6 +55,7 @@ class AugurManager:
     def __init__(self, handles_oauth=False):
         # sqlalchemy engine object
         self.engine = None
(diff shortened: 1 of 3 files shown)


cdolfi (changes requested):
Drop the pie chart changes and then lgtm. Let me know what I need to add to my local variables

JamesKunstle:
@cdolfi the changes to the piechart file are from pre-commit, I'd prefer to just accept them here rather than creating a new PR for them.

cdolfi:
@JamesKunstle that makes 100% sense merging now

cdolfi (approved):
lgtm with context of pie updates

PR #497, opened 2023-10-05
adds acme to app route

JamesKunstle:
small change to make sure that the SSL certificate handler is included in the route we expose.


cdolfi (approved):
lgtm

PR #498, opened 2023-10-05
TEMP deregisters assignment query and visualizations

JamesKunstle:
temporarily deregisters the assignment queries and their respective functionality.

when we push to main again in the future, we'll have to remember to rebase and remove this commit.


cdolfi (approved):
lgtm

PR #500, opened 2023-10-09
graph color updates

cdolfi:
this pr updates the graph color selects to optimize for contrast between colors next to each other


JamesKunstle (approved):
looks very good to me. I think the distinctions between colors has been clarified very well, and the meaning of the colors is also clear (e.g. "unassigned" == red).

PR #501, opened 2023-10-09
search bar background color update

cdolfi:
this pr updates the background of the search bar. lmk what you think

### DIFF
--- a/8Knot/assets/color.css
+++ b/8Knot/assets/color.css
@@ -72,3 +72,7 @@ blockquote p {
 .alert-warning {
     background-color: #B5B682 !important;
 }
+
+.mantine-MultiSelect-values {
+  background-color: #272b30;
+}
--- a/8Knot/assets/color.css
+++ b/8Knot/assets/color.css
@@ -74,6 +74,6 @@ blockquote p {
 }
 
 /* updates background of the dmc-multiselect in visualizations */
-.viz-values {
+.dmc-multiselect-custom {
   background-color: #272b30;
 }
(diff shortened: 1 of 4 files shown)


JamesKunstle:
I think the search bar styling just needs a net refactor. If the background is black, the rest sticks out and kinda looks "partially rendered" I guess. 

On a razors edge, I think I prefer the search bar background to be white along with the other stuff in the bar rather than having a black background with white outline and white button.

The search bar definitely sticks out as the remainder default-styled thing on the page though, but idk what the right styling is ultimately.

cdolfi:
@JamesKunstle Personally I agree on the repo search bar looks better in the prior format, but the visualization ones I think are improved by this change. Lmk what you think of those search bars specifically and we could just apply this change on those

JamesKunstle:
@cdolfi Yeah for the bot filters I'm on the same page. They look better with this styling change. I'd be down to make one overall styling change this time and create a subsequent issue to reimagine the search bar styling later.

cdolfi:
@JamesKunstle Sounds good I updated the pr to just apply to the visualizations

8Knot/assets/color.css
```diff
@@ -72,3 +72,8 @@ blockquote p {
 .alert-warning {
     background-color: #B5B682 !important;
 }
+
+/* updates background of the dmc-multiselect in visualizations */
+.viz-values {
```
JamesKunstle, line 77: I'd rename this to something descriptive like "dmc-multiselect-custom" because CSS is such a pain to interpret later out of context. Otherwise this works for me.

JamesKunstle (changes requested):

JamesKunstle (approved):
lgtm!

Issue #475, opened 2023-09-27
Queries: slice cntrib_ids to 20 characters 


cdolfi:
cntrb_ids carry a lot of unnecessary zeros once converted to string. 20 characters holds all necessary info plus buffer for larger cntrb_ids for the future 

df["cntrb_id"].str.slice(0, 20)

PR #507, opened 2023-10-18
cntrb id update

cdolfi:
this solves #475 to make all of the contributor ids across queries to be the same length and format


JamesKunstle (changes requested):
Looks good, but can you please update the commit message and PR description to describe what this change is for and why it's being made?

cdolfi:
@JamesKunstle Comment updated

JamesKunstle (approved):

PR #528, opened 2023-11-09
number resize

cdolfi:
Check out on smaller screen as well. Only one I dont love is the project velocity one having to have so many rows for inputs but I think thats the nature of it. Better than not being able to see what your values are


JamesKunstle (approved):
deployed it locally and checked it out- I think it looks sharper all around. 
I also don't love what happens with project_velocity but as you said, that might just be the price of all the parameters.

PR #539, opened 2023-11-14
shortens max job lifespan

JamesKunstle:
Jobs can live 30 minutes at most - shorter than the previous 24hr time limit. This should be a simple fix for observed bugs when a job gets stuck waiting on something that failed. Shouldn't be a replacement for inter-job communication but is a simple patch to minimize impact.


cdolfi (approved):
lgtm!

PR #545, opened 2023-11-15
re-registers user-group-query bug

JamesKunstle:
switches deregistration: re-registers user_groups_query and de-registers pr_assignment_query.

fixes user groups bug we saw on Openshift.


cdolfi (approved):
lgtm

PR #562, opened 2023-12-12
Search bar option fix and loading error

cdolfi:
As pointed out by @GregSutcliffe some directories were not getting listed. I figured out it was the ones that only had nested folders and no other files. Additional step makes sure that those directories are listed. 

Changed drop down to not be clearable to prevent reloads that were causing issues


JamesKunstle (approved):
this all makes sense

PR #583, opened 2023-12-18
fix queries to accept null timestamps

JamesKunstle:
queries check to make sure event timestamps didn't happen in the future.
this patch adds a check to accept a rows with NULL in timestamp
fields so that if an event hasn't completed (issue or PR closed)
the row is still accepted.

### DIFF
--- a/8Knot/cache_manager/db_init.py
+++ b/8Knot/cache_manager/db_init.py
@@ -117,11 +117,12 @@ def _create_application_tables() -> None:
 
     with conn.cursor() as cur:
         # create tables if they don't already exist.
+        # TODO: id->repo_id, commits->commit_id
         cur.execute(
             """
             CREATE UNLOGGED TABLE IF NOT EXISTS commits_query(
-                id integer,
-                commits text PRIMARY KEY,
+                id int,
+                commits text, -- this is the commit hash, so it's base64 hash.
                 author_email text,
                 date text,
                 author_timestamp text,
@@ -133,11 +134,11 @@ def _create_application_tables() -> None:
         cur.execute(
             """
             CREATE UNLOGGED TABLE IF NOT EXISTS issues_query(
-                id integer,
+                id int,
                 repo_name text,
-                issue text,
-                issue_number text,
-                gh_issue text,
+                issue int,
+                issue_number int,
+                gh_issue int,
                 reporter_id text,
                 issue_closer text,
                 created text,
@@ -150,10 +151,10 @@ def _create_application_tables() -> None:
         cur.execute(
             """
             CREATE UNLOGGED TABLE IF NOT EXISTS prs_query(
-                id integer,
+                id int,
                 repo_name text,
-                pull_request text,
-                pr_src_number text,
+                pull_request int,
+                pr_src_number int,
                 cntrb_id text,
                 created text,
                 closed text,
@@ -168,10 +169,10 @@ def _create_application_tables() -> None:
             CREATE UNLOGGED TABLE IF NOT EXISTS company_query(
                 cntrb_id text,
                 created text,
-                id integer,
+                id int,
                 login text,
                 action text,
-                rank integer,
+                rank int,
                 cntrb_company text,
                 email_list text
             )
@@ -182,13 +183,13 @@ def _create_application_tables() -> None:
         cur.execute(
             """
             CREATE UNLOGGED TABLE IF NOT EXISTS contributors_query(
-                id integer,
+                id int,
                 repo_name text,
                 cntrb_id text,
                 created_at text,
                 login text,
                 action text,
-                rank integer
+                rank int
             )
             """
         )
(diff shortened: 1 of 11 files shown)


8Knot/pages/codebase/visualizations/cntrb_file_heatmap.py
```diff
@@ -218,6 +218,41 @@ def directory_dropdown(repo_id):
     return directories, "Top Level Directory"
 
 
+# callback for contributor file heatmap graph
```
JamesKunstle, line 221: This is added/subtracted because we moved the multi_query_helper down below this function. Otherwise no changes have been made.

8Knot/cache_manager/db_init.py
```diff
@@ -117,11 +117,12 @@ def _create_application_tables() -> None:
...
     with conn.cursor() as cur:
         # create tables if they don't already exist.
+        # TODO: id->repo_id, commits->commit_id
         cur.execute(
             """
             CREATE UNLOGGED TABLE IF NOT EXISTS commits_query(
-                id integer,
-                commits text PRIMARY KEY,
+                id int,
+                commits text, -- this is the commit hash, so it's base64 hash.
```
cdolfi, line 125: what does this mean exactly (not a needed change, just want to make sure I understand)

8Knot/queries/issue_assignee_query.py
```diff
@@ -45,9 +45,11 @@ def issue_assignee_query(self, repos):
                         explorer_issue_assignments ia
                     WHERE
                         ia.id in %s
-                        and ia.created < now()
-                        and ia.closed < now()
-                        and ia.assign_date < now()
+                        and (ia.created < now() or ia.created IS NULL)
```
cdolfi, line 48: Created should(cant?) never be null as if it was never created it doesnt exist

8Knot/queries/issues_query.py
```diff
@@ -48,8 +48,10 @@ def issues_query(self, repos):
                         r.repo_id = i.repo_id AND
                         r.repo_id in %s
                         and i.pull_request_id is null
-                        and i.created_at < now()
-                        and i.closed_at < now()
+                        and (i.created_at < now() or i.created_at IS NULL)
```
cdolfi, line 51: same as above

8Knot/queries/pr_assignee_query.py
```diff
@@ -45,9 +45,11 @@ def pr_assignee_query(self, repos):
                         explorer_pr_assignments pa
                     WHERE
                         pa.id in %s
-                        and pa.created < now()
-                        and pa.closed < now()
-                        and pa.assign_date < now()
+                        and (pa.created < now() or pa.created IS NULL)
```
cdolfi, line 48: same as above

8Knot/queries/prs_query.py
```diff
@@ -28,7 +28,11 @@ def prs_query(self, repos):
     if len(repos) == 0:
         return None
 
-    query_string = f"""
+    # NOTE: This was dropping PRs that were needed in the heatmap
+    # visualizations, so we switched back to the old query.
```
cdolfi, line 32: wdym switched back to the old query? What was the new one exactly?

8Knot/queries/repo_files_query.py
```diff
@@ -36,27 +36,25 @@ def repo_files_query(self, repos):
...
+                        rl.rl_analysis_date,
+                        rl.file_path,
+                        rl.file_name
+                    FROM
+                        repo_labor rl,
+                        repo r
+                    WHERE
+                        rl.repo_id = r.repo_id AND
+                        rl.repo_id in %s AND
+                        (rl.repo_id, rl.rl_analysis_date) IN (
```
cdolfi, line 51: Add some comments in on what is being done here

cdolfi (changes requested):
few small things and this will be ready to go!

8Knot/cache_manager/db_init.py
JamesKunstle: Every commit has a hash value, which is unique. In the augur commits table, multiple rows have the same hash, referring to the same commit, but to different files in the commit. the hash is in 'base64' which means its a string that looks like "012a1bbed8a9092dec29fff" or something, which is a more compact form for representing binary. 

so the datatype is text, because utf-8 characters are being used to represent many binary digits.

8Knot/queries/issue_assignee_query.py
JamesKunstle: tru

8Knot/queries/issues_query.py
JamesKunstle: ++

8Knot/queries/pr_assignee_query.py
JamesKunstle: ++

8Knot/queries/prs_query.py
JamesKunstle: bad comment, will remove

8Knot/queries/repo_files_query.py
JamesKunstle: kk

cdolfi (approved):
lgtm! Squash commits and ill merge

PR #566, opened 2023-12-13
Refactor recent visualizations -> Postgres caching

JamesKunstle:
* adds tables with db_init for:
	- cntrb_per_file_query
	- pr_files_query
	- pr_response_query
	- repo_files_query
* refactors above queries to use cache_facade and write to postgres
* updates multiple visualizations to read from postgres_cache including:
	- both heatmaps
	- PR messaging response window

** I'm sorry the linter got run and committed, the major real changes are above.

PR #584, opened 2023-12-18
Dev postgres stage

JamesKunstle:
Finalizing this merge process in this PR because, with  #583  being merged, it's cleaner to do it this way. #566 uses a branch from my fork that is behind dev-postgres-stage


cdolfi (approved):
lgtm! everything looks great when I run locally

PR #588, opened 2023-12-18
points redis-users deployment to correct PVC

JamesKunstle:
regular redis cache and redis-users cache were using the same persistent volume claim. this points the redis-users instance to its own PVC


cdolfi (approved):
lgtm

PR #589, opened 2023-12-18
Add ocp postgres

JamesKunstle:
defines postgres deployment + pvc + service as part of base deployment

will have to be merged after #588

### DIFF
--- a/openshift/base/8k-app-server.yaml
+++ b/openshift/base/8k-app-server.yaml
@@ -71,6 +71,19 @@ spec:
             requests:
               cpu: 50m
               memory: 150Mi
+      initContainers:
+        # Set up Postgres cache before app server
+        # becomes available.
+        - name: postgres-initialization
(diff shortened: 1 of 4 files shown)


JamesKunstle:
@codekow would love your input here as well

JamesKunstle:
> @JamesKunstle
> 
> A few things to consider:
> 
> * You can reuse the secret for both DBs if you wanted
> * You'll need to update the `kustomization.yaml` to use the new resources
> * I like the idea of your application having the ability to detect a cache DB and if the cache isn't present, it runs without.

1. the same secret for Redis and for Postgres? Yeah sure that'd be fine. That's easier to maintain
2. okay, just did that
3. that'd be nice in the future. right now, we have an inflexible relationship between the query and the analysis workers- an architectural defect from earlier on. In the future, a cache-less mode would be possible.

PR #593, opened 2023-12-19
Codekow review

codekow:


openshift/base/secret-postgres.yaml
```diff
@@ -6,4 +6,8 @@ metadata:
     app.kubernetes.io/part-of: eightknot-app
   name: eightknot-postgres
 stringData:
-  POSTGRES_PASSWORD: somepasswordfortesting
+  # TODO: use specific vars
+  # POSTGRESQL_USER: cache
+  # POSTGRESQL_PASSWORD: somepasswordfortesting 
+  # POSTGRESQL_DATABASE: augur_cache
+  POSTGRESQL_ADMIN_PASSWORD: somepasswordfortesting
```
JamesKunstle, line 13: The postgres image, at least the enterprisedb one, expects that the POSTGRESQL_PASSWORD environment variable is available. Not sure if that's the case for the imagestream one. I know that I'm not looking for POSTGRESQL_ADMIN_PASSWORD in my environment variable checks in application logic, could switch over though.

codekow: The other image uses different vars and I had to update the secret to have both in it for now to work

JamesKunstle (approved):
Thanks Cory!

PR #596, opened 2023-12-20
patch kustomization for prod imagestream

JamesKunstle:
@codekow patching kustomization to point to correct imagestream


JamesKunstle:
This works when deployed to main so I think it's chill to merge

PR #604, opened 2024-01-04
Repo Info Page 

cdolfi:
First round pass at page with a different style of information. Not tied to any titles or page design

### DIFF
--- a/8Knot/cache_manager/db_init.py
+++ b/8Knot/cache_manager/db_init.py
@@ -263,6 +263,72 @@ def _create_application_tables() -> None:
 
         cur.execute(
             """
+            CREATE UNLOGGED TABLE IF NOT EXISTS repo_languages_query(
+                id int,
+                programming_language text,
+                code_lines int,
+                files int
+            )
+            """
+        )
+        logging.warning("CREATED repo_languages_query TABLE")
+
+        cur.execute(
+            """
+            CREATE UNLOGGED TABLE IF NOT EXISTS package_version_query(
+                id int,
+                name text,
+                current_release_date text,
+                latest_release_date text,
+                libyear float4
+            )
+            """
+        )
+        logging.warning("CREATED package_version_query TABLE")
+
+        cur.execute(
+            """
+            CREATE UNLOGGED TABLE IF NOT EXISTS repo_releases_query(
+                id int,
+                release_name text,
+                release_created_at text,
+                release_published_at text,
+                release_updated_at text
+            )
+            """
+        )
+        logging.warning("CREATED repo_releases_query TABLE")
+
+        cur.execute(
+            """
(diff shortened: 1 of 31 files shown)


8Knot/cache_manager/db_init.py
```diff
@@ -261,6 +261,72 @@ def _create_application_tables() -> None:
...
+                score float4
+            )
+            """
+        )
+        logging.warning("CREATED ossf_score_query TABLE")
+
+        cur.execute(
+            """
+            CREATE UNLOGGED TABLE IF NOT EXISTS repo_info_query(
+                id int,
```
JamesKunstle, line 316: is `id` == `repo_id` in this file? If so I think it should be named that way

8Knot/pages/index/index_layout.py
```diff
@@ -90,6 +90,7 @@
                             dbc.Nav(
                                 [
                                     dbc.NavLink("Welcome", href="/", active="exact"),
+                                    dbc.NavLink("Repo Info", href="/repo_info", active="exact"),
```
JamesKunstle, line 93: Maybe this is a good opportunity to use the term `Repo Overview` because we already use `Info` in a tab.

JamesKunstle (changes requested):
Almost everything looks great- would love 'id' to be called 'repo_id' in queries and analysis if that's what it's for, and have a titling question

8Knot/pages/index/index_layout.py
cdolfi: Good call

8Knot/cache_manager/db_init.py
cdolfi: will that break things? I didnt know if id was still a keyword or if we want to change that when we change all of them

JamesKunstle: Nah we can change it w/o a problem. In the query, I think it's `repo_id as id` so that'd just go back to `repo_id` and when you use the column in the analysis you'd just use `id` too.

8Knot/pages/index/index_layout.py
JamesKunstle: ++

8Knot/cache_manager/db_init.py
JamesKunstle: okay we'll go back and update that when we change all the names

JamesKunstle (approved):

JamesKunstle:
@cdolfi rebase this branch onto dev to fix the merge conflict and then we're gtg

cdolfi:
@JamesKunstle should be good to go

JamesKunstle:
@cdolfi weird, I'm still getting "This branch cannot be rebased due to conflicts", are you getting that or are you able to merge it yourself?

PR #608, opened 2024-01-04
welcome page plotly tab update

cdolfi:
Change title to “Using 8Knot Visualizations”
add a “# 5 About Graph” that has more details about the type of info that people find under that button, because it’s way more useful than what people might think for an about link.


JamesKunstle (approved):
very good change, ++

PR #610, opened 2024-01-04
PR and issue assignment optimization 

cdolfi:
Might need more updates to optimize more but this makes a HUGE difference

### DIFF
--- a/8Knot/pages/contributions/visualizations/cntrb_pr_assignment.py
+++ b/8Knot/pages/contributions/visualizations/cntrb_pr_assignment.py
@@ -51,6 +51,27 @@ gc_cntrib_pr_assignment = dbc.Card(
                         dbc.Row(
                             [
                                 dbc.Label(
+                                    "Date Interval:",
+                                    html_for=f"date-radio-{PAGE}-{VIZ_ID}",
+                                    width="auto",
+                                ),
+                                dbc.Col(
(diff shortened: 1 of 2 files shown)


JamesKunstle:
@cdolfi This limits the range of dates for which the graph is rendered, correct? So, we're visualizing the data the same way, but we're visualizing less data at once by default with this change.

cdolfi:
@JamesKunstle Yes precisely, the initial rendering takes so much less time than before

JamesKunstle:
@cdolfi In testing, this is much better than before. I think we still ought to aggregate small values of assignments, and let the user know that un-aggregating will have a browser performance penalty. For the Kubernetes org, I get the following:

Screenshot 2024-01-08 at 4 27 06 PM

Which is very slow to interact with.

cdolfi:
@JamesKunstle That makes sense to me, can you detail what you have in mind in the issue linked? Think itd be good to do in a separate PR (and would not close said issue w this PR)

JamesKunstle:
Yeah I definitely can

JamesKunstle:
#613 opened this issue, and #614 this one

JamesKunstle (approved):

JamesKunstle:
you'll have to rebase the branch manually for it to be merged.

PR #612, opened 2024-01-05
drive repeat error handling

cdolfi:
if no data in drive or repeat there would be an error, this update solves that issue


JamesKunstle (approved):
++ good fix

PR #626, opened 2024-01-23
make workers more resillient

JamesKunstle:
prefetch and concurrency were pinning workers to
jobs that weren't going to terminate. this should
alleviate some worker pool exhaustion by cancelling deadlocked tasks and allowing workers to immediately pick up tasks from queue if they're available.

### DIFF
--- a/8Knot/_celery.py
+++ b/8Knot/_celery.py
@@ -16,7 +16,7 @@ celery_app = Celery(
 )
 
 celery_app.conf.update(
-    task_time_limit=2700, # 45 minutes
+    task_time_limit=2700,  # 45 minutes
     task_acks_late=True,
     task_track_started=True,
     result_extended=True,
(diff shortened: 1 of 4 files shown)


JamesKunstle:
@codekow This PR makes Celery Worker daemons effectively single-task so Openshift can manage the scale-up, and we avoid tasks being reserved by workers that will block forever.

Could you please spot-check this?

JamesKunstle:
tl;dr: make workers single-task, disable prefetch, bump # workers, decrease ttl for deadlocked tasks.

Won't need query workers when app and db are on the same hardware and we can bypass caching, and the data MUST be available because we have a copy of the db, so we shouldn't have to deal w/ deadlocks in the future.

deadlocks arise because the analysis workers are polling, waiting for data to be queried from the main db and cached. if something goes wrong, they'll wait until their terminated, blocking a worker from attending to other user's requests. 

This polling behavior should be remedied by this fix: https://github.com/plotly/dash/issues/2725

PR #641, opened 2024-02-02
Project Velocity: Fix for projects with nan values

cdolfi:
Before if any issue or pr action was were nan, then the whole weighted value would be nan. This pr fixes that issue for all repositories with atleast 1 pr or issue action will be plotted


JamesKunstle (approved):
good fix!

PR #644, opened 2024-02-12
minor changes to README.md, env requirements

JamesKunstle:
@cdolfi This updates docs to reflect some changes to the environment requirements and the folder structure. I'm 50/50 on sketching out the caching system before we update the augur replica. What are your thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,24 +40,40 @@ The file-structure of our application is intuitive. Non-application files are om
     |   +-- index/
                |     +-- index_callbacks.py
                |     +-- index_layout.py
-        +-- overview/
-               |     +-- overview.py
+        +-- welcome/
+               |     +-- welcome.py
                |     +-- visualizations/
                             |     +-- name_of_visualization_1.py
                             |     ...
                             |     +-- name_of_visualization_n.py
+        +-- contributions/
+               |     +-- contributions.py
+               |     +-- visualizations/
+                            |     ...
+        +-- contributors/
+               |     +-- cntrb_behavior.py
+               |     +-- cntrb_type.py
+               |     +-- visualizations/
+                            |     ...
         +-- chaoss/
                |     +-- chaoss.py
                |     +-- visualizations/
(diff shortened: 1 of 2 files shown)


cdolfi:
@JamesKunstle I believe we should bring the docs to date to the current structure, especially with the amount of people trying 8Knot for the first time. I would say if there is portions of it you think will be changed soon I wouldnt go into too much detail

JamesKunstle:
@cdolfi Which structure are you specifically describing? The caching architecture or the file layout structure?

cdolfi:
@JamesKunstle Structure is not actually the right word to use here my b. My only concern at this point is that our readme has the updated information on how to locally host and run an 8knot instance. On the description of the caching, maybe just a little burb with a 1 sentance-ish description of what we have now and that it will be updated soon with the new changes

JamesKunstle:
@cdolfi Okay, I updated the docs to be clearer about how to configure and deploy the application, and a comment about our caching likely changing in the near future. What do you think?

README.md
```diff
@@ -120,14 +136,28 @@ If you would prefer to look at our most up-to-date work, please check out the fo
 
 ---
 
-## Local Development
+## Architectural Comment
 
-We've tried to make it as easy as possible to go from fork/clone to a running instance.
+Our demonstration application run in us-west2, while the Augur database that we use is in Columbus, MO. To improve user experience,
```
cdolfi, line 141: Outside of Red Hat I dont think people will know what "us-west2" means

cdolfi (changes requested):
one small thing and then itll be good to go from my side!

README.md
JamesKunstle: Oh, that's an AWS zone. I'll clarify but that's a conventional AWS region that's well-known.

JamesKunstle: updated!

cdolfi (approved):
lgtm

PR #648, opened 2024-02-16
🚚 Update URIs to current GitHub repository

jwflory:
While seeking out how to report a bug in the project, I realized that the old repository/project name was still used in several places. This commit updates all remaining references from `sandiego-rh` to `oss-aspen`, matching the new name and pointing users the right way.


cdolfi (approved):
lgtm!

PR #652, opened 2024-03-01
Heatmap function breakdown and change the df_actions to be based on search bar results

cdolfi:


JamesKunstle (changes requested):
what's the rationale of changing the use of the repolist for the cache querying? otherwise lgtm

cdolfi:
@JamesKunstle Its changing to repolist since the dates are now based on the search bar instead of the 1 repo. I could be misunderstanding how to use caching in this scenario with it being for a single repo but dates based on all. Lmk if this doesnt make sense and we can talk

JamesKunstle (approved):

PR #664, opened 2024-03-12
pie update

cdolfi:
Fix for the lottery factor chart. Once this one is agreed on Ill fix the other instance of this viz


JamesKunstle (approved):
seems like you had a clear fix, good w/ me

PR #683, opened 2024-04-03
error handle when there is no assignment data

cdolfi:
Examples where you can see behavior: 
https://github.com/Audacium/audacium/
https://github.com/Quagga/quagga/

The original implementation of error handing was faulty, and even so did not handle for date range. This solves for both


JamesKunstle (approved):
++ good fix

PR #688, opened 2024-04-05
Lottery Factor Line Graph: Pattern and date range removal

cdolfi:
This graph did/does not have the same bug that the pie chart had. During this audit it was discovered that the date range functionality was broken. After viewing this graph with a project with a 10+ year history, I found this feature unnecessary. In this pr I removed the date picker for reasons above and the pattern selector as that is prior implementation of github bots. Old viz for reference:
Screenshot 2024-04-05 at 1 30 55 PM

### DIFF
--- a/8Knot/pages/contributors/visualizations/contrib_importance_over_time.py
+++ b/8Knot/pages/contributors/visualizations/contrib_importance_over_time.py
@@ -65,29 +65,6 @@ gc_lottery_factor_over_time = dbc.Card(
                         dbc.Row(
                             [
                                 dbc.Label(
-                                    "Threshold:",
-                                    html_for=f"threshold-{PAGE}-{VIZ_ID}",
-                                    width="auto",
-                                ),
-                                dbc.Col(
-                                    [
-                                        dcc.Slider(
-                                            id=f"threshold-{PAGE}-{VIZ_ID}",
-                                            min=10,
-                                            max=95,
-                                            value=50,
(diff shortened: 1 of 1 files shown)


JamesKunstle:
What specifically are you removing here? I pulled your branch and it still has the date-picker but not the threshold range-slider or the bot pattern matcher.

cdolfi:
@JamesKunstle let me go check when I get back to my computer. It should be removing the date range picker and patterns so something might have gotten messed up when I pushed

cdolfi:
@JamesKunstle This is what I see when I run it locally. Let me know if you are seeing something different 
Screenshot 2024-04-09 at 4 56 18 PM

JamesKunstle:
@cdolfi Yeah that's not what I'm seeing- lemme make sure that I fully rebuild and everything

JamesKunstle:
@cdolfi I confirmed that I'm starting 8Knot with the commit that's top of this branch, could you confirm that you've pushed all of your changes?

Screenshot 2024-04-10 at 12 06 27 PM

cdolfi:
@JamesKunstle This PR is for the Lottery Factor line graph not the project velocity visualization

JamesKunstle:
@cdolfi My bad, I swapped the two in my head and was looking at the date picker / value adjustment boxes. I confirmed that the lottery factor viz renders and behaves as expected on my end.

PR #690, opened 2024-04-05
project velocity viz log 0 error handling

cdolfi:
Error was happening when there were 0 commits (same would happen to contributors). This fix prevents log from being performed on 0 values. More details in issue


JamesKunstle (approved):
v good change

PR #697, opened 2024-04-12
Codebase page: swap first 2 heatmaps

cdolfi:
I have been thinking about this for a while. Swapping order to put the most conceptually accessible graph first and the contributor/reviewer one next to each other


JamesKunstle (approved):
lgtm

PR #711, opened 2024-05-09
graph button fix

### LINKED ISSUE #670: Fix "about graph" button location on response graphs
On smaller screens the prior implementation would have the button and input on different lines

cdolfi:
On smaller screens the prior implementation would have the button and input on different lines, closes #670


JamesKunstle (approved):
This is just a styling change to I don't see a reason not to merge it if it looks good to you.

px is preferred for width as opposed to rm or em.

PR #712, opened 2024-05-09
normalized email case

### LINKED ISSUE #672: Contributor email case is not normalized
**Describe the bug**
Contributor email case is not normalized. For example, there separate slices for `Gmail.com` and `gmail.com`.

RFC 1035: Domain names - implementation and specification: [2.3.1. Preferred name syntax](https://www.rfc-editor.org/rfc/rfc1035.html#section-2.3.1):
> Note that while upper and lower case letters are allowed in domain names, no significance is attached to the case.  That is, two names with the same spelling but different case are to be treated as if identical.

**To Reproduce**
1. Launch docker (per Insight out of CHAOSS: Community Data Analysis Tool Workshop)
2. Click on Affiliation (http://127.0.0.1:8080/affiliation)

**Expected behavior**
Data should be normalized so that there is a single slice for `gmail.com`.

**Screenshots**
Screenshot 2024-03-14 at 11 15 39

**Additional context**
Fantastic presentation! Thank you ❤️

cdolfi:
change all emails to lowercase for better grouping, closes #672

### DIFF
--- a/8Knot/pages/affiliation/visualizations/org_associated_activity.py
+++ b/8Knot/pages/affiliation/visualizations/org_associated_activity.py
@@ -221,8 +221,8 @@ def process_data(df: pd.DataFrame, num, start_date, end_date, email_filter):
     # creates list of emails for each contribution and flattens list result
     emails = df.email_list.str.split(" , ").explode("email_list").tolist()
 
-    # remove any entries not in email format
-    emails = [x for x in emails if "@" in x]
+    # remove any entries not in email format and flattens list result
+    emails = [x.lower() for x in emails if "@" in x]
 
     # creates list of email domains from the emails list
     email_domains = [x[x.rindex("@") + 1 :] for x in emails]
--- a/8Knot/pages/affiliation/visualizations/commit_domains.py
+++ b/8Knot/pages/affiliation/visualizations/commit_domains.py
@@ -178,8 +178,8 @@ def process_data(df: pd.DataFrame, num, start_date, end_date):
     # creates list of emails for each contribution and flattens list result
     emails = df.author_email.tolist()
 
-    # remove any entries not in email format
-    emails = [x for x in emails if "@" in x]
+    # remove any entries not in email format and put all emails in lowercase
+    emails = [x.lower() for x in emails if "@" in x]
 
     # creates list of email domains from the emails list
     email_domains = [x[x.rindex("@") + 1 :] for x in emails]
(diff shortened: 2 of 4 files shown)


8Knot/pages/affiliation/visualizations/org_associated_activity.py
```diff
@@ -221,8 +221,8 @@ def process_data(df: pd.DataFrame, num, start_date, end_date, email_filter):
     # creates list of emails for each contribution and flattens list result
     emails = df.email_list.str.split(" , ").explode("email_list").tolist()
 
-    # remove any entries not in email format
-    emails = [x for x in emails if "@" in x]
+    # remove any entries not in email format and flattens list result
+    emails = [x.lower() for x in emails if "@" in x]
```
JamesKunstle, line 225: noticing this for the first time- might prefer to use a regular expression for email correctness, rather than just checking for the `@` sign.

8Knot/pages/affiliation/visualizations/org_associated_activity.py
```diff
@@ -221,8 +221,8 @@ def process_data(df: pd.DataFrame, num, start_date, end_date, email_filter):
     # creates list of emails for each contribution and flattens list result
     emails = df.email_list.str.split(" , ").explode("email_list").tolist()
 
-    # remove any entries not in email format
-    emails = [x for x in emails if "@" in x]
+    # remove any entries not in email format and flattens list result
```
JamesKunstle, line 224: what does 'flatten list result' mean?

JamesKunstle (changes requested):
generally lgtm, just want to clarify comments and check if we want to use a regex rather than just using the @ symbol to check for emails.

8Knot/pages/affiliation/visualizations/org_associated_activity.py
cdolfi: What do you see the regex providing that this solution does not?

JamesKunstle: if we're seeing "emails" that aren't actually email format, we can use the canonical "is this an email" regex to check for the real email rules. If we don't really care about that level of formalism, no need to pursue it.

8Knot/pages/affiliation/visualizations/org_associated_activity.py
JamesKunstle: ++

8Knot/pages/affiliation/visualizations/org_associated_activity.py
cdolfi: Yeah I think at the current state that level of formalism is above what we are looking for

JamesKunstle (approved):

Issue #746, opened 2025-03-19
Affiliation Page: switch order of visualizations for its consistent of contributor vs contribution count


cdolfi:
Visualization order: 

Commit Activity by Domain                  Unique Contributor Email Domains 
Organization Associated Activity         Organization Core Contributors

PR #753, opened 2025-04-08
Affiliation Page Update - Solving Issue #746

EngCaioFonseca:
The PR solves the #746 - Affiliation Page: switch order of visualizations for its consistent of contributor vs contribution count

Tested on local instance.

Signed: Caio Fonseca


cdolfi (changes requested):
Same as the other PRs, squash commits and only include the changes that impact this specific PR

PR #758, opened 2025-04-17
Switching Python Pkg Manager (from Pip to UV) to increase Docker Building speeds - More efficiency

EngCaioFonseca:
UV as a Python Pkg Manager is growing in popularity due to its high speed in installing dependencies and its efficient commands, making it easier to transition from pip to UV. 
The UV has been added into the Dockerfile, and it install all dependencies accordingly. 
The build has been tested in a local instance, and benchmarks were compared between the UV version and the Pip version. The UV version reaches double the speed, so the build, is on average, done in half the time.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,10 +2,13 @@ FROM registry.access.redhat.com/ubi9/python-39:latest
 
 WORKDIR /opt/app-root/src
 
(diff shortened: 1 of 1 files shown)


JamesKunstle (approved):
UV is rad, I see no problem preferring it.

PR #766, opened 2025-04-29
Search bar features

EngCaioFonseca:
Search bar feature implementations: 

Debounce modification from 100 ms to 300 ms;

Prefix implementation: "repo:" and "org" before the names;

Fuzzy search with Levenshtein distance algorithm; (Still test phase)

Client side pre-caching. (still test phase).


EngCaioFonseca:
Closed as it is updated and being resolved by #777

Issue #765, opened 2025-04-28
WORKER Timeouts


sgoggins:
**Describe the bug**
After a period of time running 8Knot on a server there is eventually an issue that seems unresolvable that occurs, leading to a "WORKER TIMEOUT" that is unresolvable. 

**To Reproduce**
1. Ubuntu 22.x
2. Augur on the same server
3. run 8Knot for a while with nginx public facing page
4. Eventually these errors occur and do not seem to be resolvable 

```bash
app-server-1       | 2025-04-28 12:48:55,027 CRITICAL somepassword
app-server-1       | 2025-04-28 12:48:55,574 WARNING  LOGIN DISABLED
app-server-1       | 2025-04-28 12:48:55,591 WARNING  AUGUR: Connection to DB succeeded
app-server-1       | [2025-04-28 12:49:21 +0000] [1] [CRITICAL] WORKER TIMEOUT (pid:72)
app-server-1       | [2025-04-28 12:49:22 +0000] [1] [ERROR] Worker (pid:72) was sent SIGKILL! Perhaps out of memory?
app-server-1       | [2025-04-28 12:49:22 +0000] [137] [INFO] Booting worker with pid: 137
app-server-1       | 2025-04-28 12:49:24,289 WARNING  AUGUR: Connection to DB succeeded
app-server-1       | 2025-04-28 12:49:24,289 WARNING  MULTISELECT_STARTUP
app-server-1       | 2025-04-28 12:49:26,556 WARNING  MULTISELECT_QUERY
app-server-1       | 2025-04-28 12:49:27,316 WARNING  MULTISELECT_FINISHED
app-server-1       | 2025-04-28 12:49:27,340 CRITICAL somepassword
app-server-1       | 2025-04-28 12:49:28,105 WARNING  LOGIN DISABLED
app-server-1       | 2025-04-28 12:49:28,131 WARNING  AUGUR: Connection to DB succeeded
```

sgoggins:
My Theory is that the postgresql 16 container has changed the default local authentication from MD5 to scrhram-256

cdolfi:
If you change the docker-compose line 15 to `restart: always` it will actually initialized. This IS NOT a fix just an ability to understand the behavior

cdolfi:
<img width="1376" alt="Image" src="https://github.com/user-attachments/assets/3c7fe6f5-9cbf-4e97-8126-ccf810fa9e50" />

PR #770, opened 2025-05-02
db_Init issue fixed

EngCaioFonseca:
This PR fixes the db_init issue as presented in #765 WORKER Timeouts.

In the docker-compose file, there was already a depends-on condition for the db_init to start only after postgres was ready, however, it was not making sure that postgres was actually initialized before the db was initialized. 
This PR ensures that postgres is correctly initialized and pass a health check before the db is initialized.

It also creates a function connection with retries, where it'll go through the same process of connecting the db, but if it fails, it will try the connection again for a max number of retries.

### DIFF
--- a/8Knot/cache_manager/cx_common.py
+++ b/8Knot/cache_manager/cx_common.py
@@ -3,6 +3,7 @@ Connection Common file - accessing environment variables
 """
 import os
 import logging
+import time
 
 # credentials to access database from environment
 try:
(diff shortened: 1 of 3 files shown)


cdolfi:
LGTM, giving some time for @JamesKunstle to take a look before approving and merging

cdolfi (changes requested):
squash commits and ill merge

JamesKunstle (approved):
This is a reasonable solution.

The initial solution is to let the docker-compose mechanics handle the retry rather than retry in the db_init script directly but if this helps solve the problem then I'm for it.

cdolfi (approved):

Issue #764, opened 2025-04-25
Investigate low memory issue further


cdolfi:
<img width="1376" alt="Image" src="https://github.com/user-attachments/assets/6231747a-db2a-4beb-849e-374151c9efae" />

PR #779, opened 2025-05-21
Investigation and Proposed solution to Redis low memory issue #764

EngCaioFonseca:
Common Redis warning caused due to memory overcommit settings in Linux. 

- Linux has a feature called "memory overcommit" that affects how processes reserve memory; 
- When Redis saves data to disk or replicates, it needs to fork processes. If memory overcommit is disabled, these forks can fail when memory is tight.

The impact of this are: Redis might fail to save data to disk when memory is low. Replication might break, and these failures can cause data loss in certain situations.

Proposed solution for the Linux system is described on the code modifications in this PR. 
It is critical for production, in order to prevent data loss. For development, it is not, as it's generally a warning that things might go wrong in certain situations. For macOS development, for example, it is informing that the system has a conservative memory overcommit policy, which can mean that under very high memory pressure, the Redis background operations might fail, but for development work, it won't be an issue most times.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,7 @@ services:
     # so it'll fail. We want this to restart when it fails, because
     # postgres will be ready eventually. we don't want it to restart
     # after it succeeds. give it 1000 attempts I guess.
-    restart: on-failure:1000 
+    restart: on-failure:1000
 
   reverse-proxy:
     image: nginx:latest
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -101,6 +101,11 @@ services:
     env_file:
       - ./env.list
     restart: always
+    # NOTE: This setting is for Linux production environments.
+    # It will cause errors on macOS/Windows Docker Desktop, but is kept
+    # in the configuration for deployment to Linux servers.
+    sysctls:
+      - vm.overcommit_memory=1
 
   # for user session storage
   redis-users:
@@ -116,6 +121,11 @@ services:
     env_file:
       - ./env.list
     restart: always
+    # NOTE: This setting is for Linux production environments.
+    # It will cause errors on macOS/Windows Docker Desktop, but is kept
+    # in the configuration for deployment to Linux servers.
+    sysctls:
+      - vm.overcommit_memory=1
 
   postgres-cache:
     image: docker.io/library/postgres:16


MoralCode:
Apparrently this is not something that podman allows. Enabling the setting per the comments leads to `Error: sysctl 'vm.overcommit_memory' is not allowed` errors when starting the container:

```
--> ca48ccef3069
Successfully tagged localhost/8knot_app-server:latest
Successfully tagged localhost/8knot_worker-callback:latest
Successfully tagged localhost/8knot_worker-query:latest
Successfully tagged localhost/8knot_db-init:latest
ca48ccef30697f3aa241e59be40d6681f8dde33842dd7a0e1bef83776957c788
ca48ccef30697f3aa241e59be40d6681f8dde33842dd7a0e1bef83776957c788
ebf8fd43078107265450b059ee54f59f58ea92b4643f59513f63c1475c07366e
Error: sysctl 'vm.overcommit_memory' is not allowed
Error: sysctl 'vm.overcommit_memory' is not allowed
17bc07ea6ce9dad66015334aedd64d2e51108e2aafc43e8b189e44a7721d8fa2
917ab0d9eb6347230fd9581c25ff206fcf71b0cb0dea19ec324b431834e19058
Error: "8knot_redis-cache_1" is not a valid container, cannot be used as a dependency: no container with name or ID "8knot_redis-cache_1" found: no such container
Error: "8knot_redis-users_1" is not a valid container, cannot be used as a dependency: no container with name or ID "8knot_redis-users_1" found: no such container
Error: "8knot_redis-users_1" is not a valid container, cannot be used as a dependency: no container with name or ID "8knot_redis-users_1" found: no such container
Error: "8knot_redis-users_1" is not a valid container, cannot be used as a dependency: no container with name or ID "8knot_redis-users_1" found: no such container
```
[This](https://stackoverflow.com/a/77345711) seems to suggest that this change is something that the system administrator needs to apply to their whole system (and I assume that podman/docker simply passes the system setting into the container environment). 

running `sudo sysctl -w vm.overcommit_memory=1` to temporarily enable this, and then bringing up the compose stack with this PR as-is (i.e. with the sysctl parts commented out), redis seems to start up normally without the overcommit warning.


TL;DR a better place for this PR might be as an addition to any documentation that exists for how to run 8knot in production.

cdolfi:
@MoralCode personally Id like the warning to be cleared when running in dev as well, we can talk more about this in our meeting today

MoralCode:
I don't think this is something we can fix in the docker compose as suggested (when I tried, docker didn't do anything because this is not one of the sysctls that you can change in docker/podman due to it not being namespaced like other things are). This needs to be changed on the host system (meaning each production server this is run with, and/or each person's individual development machine)

I think the permanent home for this is in some kind of "guidance for deploying 8knot in production" document (and probably also the scripts and processes for deploying our prod instance).

cdolfi:
@MoralCode Sweet so it seems like maybe this PR just needs to change to just a docs change?

REDIS_PRODUCTION.md
```diff
@@ -0,0 +1,33 @@
+# Redis Memory Configuration for Production
```
MoralCode, line 1: I like the idea of this document overall, but maybe we can make the top level title and filename into a more general purpose `PRODUCTION_GUIDE.md`/ "guide for runing 8Knotin production" or something and make this a second level subheading or something

REDIS_PRODUCTION.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Linux Production Deployment
+
+For production deployments on Linux, you can:
+
+### Option 1: Use the production Docker Compose override file
+```bash
+docker-compose -f docker-compose.yml -f docker-compose.prod.yml up -d
+```
+
```
MoralCode, line 20: This option won't work. Docker compose isn't able to set this value based on my testing.

If we still wanted an "option 1" we could mention the temporary commands that I listed in the discussion on this issue so we have an "option 1, temporary fix for current session" and "option 2: set up your system to permanently set this setting"

docker-compose.yml
```diff
@@ -101,6 +101,12 @@ services:
     env_file:
       - ./env.list
     restart: always
+    # PRODUCTION NOTE: For Linux production environments, enable the following:
+    # sysctls:
+    #   - vm.overcommit_memory=1
+    # This prevents Redis memory overcommit warnings and potential issues.
+    # For local development on macOS/Windows, keep this commented out
+    # to avoid Docker errors.
```
MoralCode, line 109: docker doesnt allow this sysctl to be set in this way. I think we should remove these comments so we don't mislead anyone

MoralCode (changes requested):
Overall I like the idea of Caios `REDIS_PRODUCTION.md` document (although I think we should make it a more general "tips for hosting 8knot in prod" document so there's room to add on later if we discover anything else that's needed.

Details in review comments under specific code

cdolfi:
@EngCaioFonseca Can you drop the changes to the compose?

EngCaioFonseca:
@cdolfi Sure, Done.

cdolfi (approved):
lgtm!

Issue #760, opened 2025-04-17
[CI] build docker compose for every PR and merge to main


JamesKunstle:
Set up an automation that checks out this project and runs the typical build workflow.
This will catch build-time errors and bad dependency sets much earlier in development.

JamesKunstle:
Create a github action that checks out any opened PRs and runs 'docker compose up --build', then 'curl' one of the visualization-heavy pages to make sure that they run. You could instrument the backend code during testing to cause the app to fail loudly if there's an error while rendering.

PR #782, opened 2025-05-27
Add CI Automation Docker tests - Resolving issues #760

EngCaioFonseca:
The objective of this PR is to implement an automation workflow that checks if the project runs for every new PR.

This PR resolves issues described at #760.

### DIFF
--- a/8Knot/testing_utils.py
+++ b/8Knot/testing_utils.py
@@ -1,0 +1,140 @@
+"""
+Testing utilities for 8Knot application.
+
+This module provides utilities to instrument the application for better
+error detection during CI/CD testing.
+"""
+
+import os
+import sys
+import logging
+import traceback
+from functools import wraps
+
+# Configure logging for testing
+def setup_testing_logging():
+    """Configure logging to be more descriptive during testing."""
+    if os.getenv("8KNOT_DEBUG", "False") == "True":
+        logging.basicConfig(
+            level=logging.DEBUG,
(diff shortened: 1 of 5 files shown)


8Knot/testing_utils.py
```diff
@@ -0,0 +1,140 @@
...
+    checks = check_required_services()
+    for check in checks:
+        if "failed" in check.lower() or "missing" in check.lower():
+            logging.error(check)
+        else:
+            logging.info(check)
+    logging.info("=== End Service Status ===")
+
+
+def validate_visualization_data(data, visualization_name):
```
cdolfi, line 117: What exactly is this testing on? does it run just for the default search option?

cdolfi (commented):
One question and then I am good with merging this with the idea that the next sprint or two we will be intentionally looking at the results from PRs and making necessary changes

cdolfi (changes requested):
One question and one test that needs to be passed

8Knot/testing_utils.py
EngCaioFonseca: At the moment, most of these functions are not being called in the main app.py, so they are basically just created/defined, but not called. Their purpose is to be called for enhanced logging if we need it. In the case of setting the debug mode to true, we would get the data from the cache and set it to a pandas df to then send it to this function. This would be the data variable.

I think for the purpose of this PR, it is better to take all these testing utils out to a separate enhanced logging PR, and keep only the automation docker builds.

Issue #761, opened 2025-04-17
unpin dependencies on main branch


JamesKunstle:
Conventionally, it's easiest to manage dependencies when they're unpinned and you have CI running nightly to check for new versions, which can then fail freely.
When deploying a release candidate of the application, you freeze the dependencies and test the app before deploying.

cdolfi:
Discussion here: how should we be managing our dependancies? Does uv do the job? are we properly using uv? Is there an out of the box solution to this?

MoralCode:
> how should we be managing our dependancies? Does uv do the job? 

i think these days using a tool like `uv` or similar to manage dependencies is the standard practice for python projects

> are we properly using uv? Is there an out of the box solution to this?

IMO, no we aren't. It looks like uv is currently only being used to install dependencies from the `requirements.txt` files when we could be using it for development and generally making dependency management a lot easier because management tools like `uv` generate lockfiles that allow an identical set of primary and transitive dependencies to be installed across multiple machines.

I think it would help if we replaced the requirements.txt files with the `pyproject.toml` that seems to be a fairly standard format (and also what is created when using uv to initialize a new project).

@EngCaioFonseca would you be opposed to me doing a more complete migration over to `uv`? Part of me wonders if having that lockfile may help you with some of the errors you are seeing on macos. That said it may also interfere with the rewrite PR so maybe we should do this after?

cdolfi:
@MoralCode @EngCaioFonseca I will close the current PR and lets satisfy this issue with uv. What rewrite PR are you concerned with impacting?

MoralCode:
> What rewrite PR are you concerned with impacting?

I guess the impact won't be too crazy,  but I was thinking of #801

PR #786, opened 2025-05-28
Add CI Nightly dependency test - Resolving issue #761

EngCaioFonseca:
The objective of this PR is to implement a daily automation workflow that checks if the project builds if installing with the latest dependencies available.

This PR resolves issues described at #761

### DIFF
--- a/.github/workflows/nightly-dependency-test.yml
+++ b/.github/workflows/nightly-dependency-test.yml
@@ -1,0 +1,258 @@
+name: Nightly Dependency Test
+
+on:
+  schedule:
+    # Run every night at 2 AM UTC
+    - cron: '0 2 * * *'
+  workflow_dispatch: # Allow manual triggering
+
+env:
+  # Test environment variables - using public demo credentials
+  AUGUR_DATABASE: graphish
+  AUGUR_HOST: chaoss.tv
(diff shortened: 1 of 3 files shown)


MoralCode:
Overall, I'd also be curious if there are reasons to do it via a CI job (vs something like [RenovateBot](https://www.mend.io/renovate/) or one of many other similar off the shelf things). Maybe i missed the earlier design/strategy conversations and this shouldn't be the kind of thing to bring up in code review though.

cdolfi:
@MoralCode Hasnt been any other discussions beyond the issue James opened. We do use UV #758 , maybe the discussion needs to be on how to properly use uv.   I would be very pro using something off the shelf that is open source, the link to renovatebot leads to a product page fyi. Lets talk about this on the issue

MoralCode:
Done #820

cdolfi (changes requested):
please run precommit, thanks!

EngCaioFonseca:
Not sure why, but this pre-commit keeps failing. I have installed and run the pre-commit from the beginning, and locally it is passing, even manual changes are not helping, because once I commit, it brings it back to the failed state. We'll have to come back to it after the other PRs.

cdolfi (approved):
lgtm, lets keep a close eye on this a see how it runs

PR #788, opened 2025-05-28
Worker Timeout - SigKill Fix

EngCaioFonseca:
This PR increases the time of the celery workers to fix the logging warning error of worker timeout - Sigkill.


cdolfi (approved):
lgtm

PR #791, opened 2025-06-05
Add Graph Recommendation Engine

yyualvin:
## Demo Video
https://github.com/user-attachments/assets/cd969f06-2e60-4e73-ad31-23eeb8604bdc

## How this Works
Graph names and their descriptions are encoded as vectors in a Milvus vector database. When a user enters a query, a vector search will be initiated on the database, and the top 5 results are returned.

Graphs must be specified in graphs.json in this exact format.
```json
{
        "name" : "File Language By File",
        "about" : "Visualizes the percent of files or lines of code by language.",
        "id" : "gc_code_language"
},
```

## Deployment
You must specify these environmental variables in the env.list file (the embeddings model endpoint, and the credentials). Please do not use a different endpoint.

```
NOMIC_API_KEY=<Secret API Key>
NOMIC_URL=https://nomic-embed-text-v1-5-maas-apicast-production.apps.prod.rhoai.rh-aiservices-bu.com/v1/embeddings
```


8Knot/prompt.md
```diff
@@ -0,0 +1,13 @@
+## Background
```
cdolfi, line 1: what is the purpose of this file and why is it where it is?

docker-compose.yml
```diff
@@ -59,6 +59,12 @@ services:
       - EIGHTKNOT_SEARCHBAR_OPTS_MAX_RESULTS=5500  # Maximum number of results to return (org and repo)
```
cdolfi, line 59: Can you provide context to these changes?

docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/python-39:latest
+FROM registry.access.redhat.com/ubi9/python-311:latest
```
cdolfi, line 1: Does it need to be at python 11? If so we will need to check to make sure everything else is compatible with this change

requirements.txt
```diff
@@ -2,60 +2,108 @@
 # DON'T ADD NEW DEPENDENCIES TO REQUIREMENTS.TXT MANUALLY
 # ADD THEM TO REQUIREMENTS-BASE.TXT, THEN RUN
 # pip3 freeze -r requirement-base.txt > requirements.txt
```
cdolfi, line 4: @EngCaioFonseca Is this still accurate post moving over to uv?

cdolfi (changes requested):
please squash commits down, I have a few questions, and work with @EngCaioFonseca on if these changes would be compatible with a production environment and if not what needs to be iterated on

docker/Dockerfile
yyualvin: I believe llamastack requires a Python version higher than 3.9, however I can revert it back as llamastack isn't being used currently.

8Knot/prompt.md
yyualvin: That was the prompt file for when I was using an LLM to select the graphs. I'll remove this since now I'm using an embeddings model.

cdolfi:
@yyualvin Please add instructions to this on how to generate an API key and further description on how this works

PR #810, opened 2025-06-26
Add example env to repo

### LINKED ISSUE #774: Add an example .env for reference

MoralCode:
Here's a first pass at a sample .env taken largely from the README and @EngCaioFonseca's env file

TODO:
- [ ] update readme setup docs to mention using this file
- [ ] possibly update the "you have no env file" error message to reference this
- [ ] possibly provide some sensible defaults

Fixes #774


env.list.sample
```diff
@@ -0,0 +1,15 @@
+AUGUR_DATABASE=
```
cdolfi, line 1: put in the values from the readme

cdolfi (changes requested):
one small change and squash commits!

env.list.sample
MoralCode: Should I also simplify the instructions in the readme now that this file already exists? (i.e. remove the duplicate copy of the credentials and maybe just say "copy the example file and rename it to `.env` to use the default/public credentials"?

cdolfi: yeah that sounds good!

MoralCode: Done

cdolfi (approved):
lgtm

PR #818, opened 2025-07-01
Removing querying all repos from startup

EngCaioFonseca:
This PR aims to have a faster startup, by removing the synchronous querying of repos, and switching to an asynchronous query once the UI has already been started.


EngCaioFonseca:
Closing because of #858

PR #821, opened 2025-07-02
fix for repo info while more in-depth look at files query is needed

cdolfi:


MoralCode:
Its not the style of commenting things out I would have used but it does the same job.

Havent had a chance to test but the code looks good enough for what seems intended as a temporary fix

MoralCode:
taking a look at this now, but it will probably also need a rebase  too

MoralCode (approved):
LGTM

PR #822, opened 2025-07-03
Migrate from "env.list" to ".env"

### LINKED ISSUE #812: Use a .env file instead of "env.list"
env.list is not recognized as an environment file by a lot of IDEs

MoralCode:
Simple pr that fixes #812


EngCaioFonseca:
Lgtm!

PR #836, opened 2025-07-08
Nightly dependency workflow logs error build fix

EngCaioFonseca:
This PR fix the error/break on the current build with latest dependencies installed as shown by the Nightly dependency workflow. 

The error consisted in the 'dash-mantime-components==0.12.1' which after unpinned and upgraded to version 2.1.0 did not have previous keywords as used in the previous version (dropdownposition, transitiondurantion, nothingfound)


cdolfi (approved):
lgtm

PR #852, opened 2025-07-16
Removing critical info/passwords from the logs

EngCaioFonseca:
This PR removes critical info/passwords from being displayed into the logs. Resolving issue #795


EngCaioFonseca:
> In addition to this fix, I attempted to find other instances of logging sensitive information by 8Knot and wasn't easily able to find much.
> 
> This included:
> 
> * searching the codebase for instances of words like "password" "token" and "key" and looking for lines that indicated they were printed
> * searching logs of some long-running 8Knot instances for any instances of the passwords that are in use for those instances.
> 
> This appears to be the only place in the code where such passwords are logged.

Yeah, same for me. I looked through the codebase, and this seems to be the only place logging the passwords.

MoralCode (approved):

cdolfi (approved):
lgtm

Issue #845, opened 2025-07-11
Major dependency refresh


cdolfi:
To allow for an automatic dependency checking process to start out on the right foot, the dependancies in general need a major refresh so everything is as up to date as things can be where they are compatible

PR #857, opened 2025-07-16
Fix uv issues

### LINKED ISSUE #855: UV to not use a virtual environment inside the docker compose
Looks like its related to UV and virtual environment locations.

Its probably pretty benign for now, but if we wanted to It might be worth making a new PR later to get UV to not use a virtual environment inside the Docker compose Since it's already a well-controlled environment.

_Originally posted by @MoralCode in https://github.com/oss-aspen/8Knot/issues/853#issuecomment-3079374419_

### LINKED ISSUE #856: uv permissions issue at runtime
@yyualvin encountered this when running with the new `uv` setup

```
db-init-1          | error: failed to open file `/opt/app-root/src/.cache/uv/sdists-v9/.git`: Permission denied (os error 13)
app-server-1       | error: failed to open file `/opt/app-root/src/.cache/uv/sdists-v9/.git`: Permission denied (os error 13)
```

This shouldn't be happening for several reasons, but mainly because uv shouldn't be updating the cache or dependencies at runtime - they should be fixed to whatever they were at container build time.

Will make a PR for this

MoralCode:
This updates the uv config to be a little more proper (and also less broken) by essentially refactoring the whole dockerfile. This is mostly because that was the [first template i found](https://depot.dev/docs/container-builds/how-to-guides/optimal-dockerfiles/python-uv-dockerfile), but theres technical reasons for it

Changes:
- Refactored the docker compose to use a builder pattern, potentially saving disk space in the final built container
- remove `uv run` from all the various commands in the docker compose and replace it with imply adding the uv-installed binaries (like gunicorn) to PATH - this was why uv was trying to update cache files in the first place at runtime
- set the PYTHONPATH in the container to prefer dependencies from uv
- removed dependencies ` "uuid==1.30",` and `"datetime==5.4",` - they shadow python builtins and are therefore essentially unused. The above PYTHONPATH changes revealed this to be an issue during testing. While not intended, this does slightly help out the efforts of #845.
- Installs dependencies before copying the app files to make use of docker layering and caching when repeatedly building containers during development

App runs and functions fine (ignoring my janky podman) - this should, theoretically, be entirely transparent

fixes #856 and fixes #855

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ services:
     build:
       context: .
       dockerfile: ./docker/Dockerfile
-    command: ["uv", "run", "python3", "./cache_manager/db_init.py"]
+    command: ["python3", "./cache_manager/db_init.py"]
     depends_on:
       postgres-cache:
         condition: service_healthy  # to ensure that postgres is indeed initialized before this starts
@@ -30,8 +30,6 @@ services:
       dockerfile: ./docker/Dockerfile
     command:
       [
-        "uv",
-        "run",
         "gunicorn",
         "--reload", # for Production, disable the "--reload" option.
         "--bind",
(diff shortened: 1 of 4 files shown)


cdolfi (approved):
lgtm

Issue #816, opened 2025-07-01
No longer collect on 200 repositories on start up


cdolfi:

cdolfi:
long longer necessary with infra improvements

PR #858, opened 2025-07-21
Removal of pre-querying db, faster start up

EngCaioFonseca:
This PR removes the pre-querying of all repos and orgs at start up of 8knot, while still loads Chaoss org as default and gets the data ready for the user. 

The DB is only queried based on an on-demand system when the user clicks on search.

Resolves: #816


8Knot/app.py
```diff
@@ -44,8 +44,12 @@
 except SQLAlchemyError:
     sys.exit(1)
 
+# STARTUP OPTIMIZATION: multiselect_startup() disabled for significantly faster startup
```
cdolfi, line 47: Curious to hear @MoralCode thoughts on this, but I think it would be sufficient to add this in the PR description to refer back to but to remove from codebase. Without the context of the past, seems to be logical that at start up the only repo(s) loaded are the ones listed in the start up value

8Knot/cache_manager/cache_facade.py
```diff
@@ -166,9 +172,17 @@ def caching_wrapper(func_name: str, query: str, repolist: list[int], n_repolist_
     Raises:
         Exception: If a step fails, will print exception and re-raise.
 
-    Returns:
+        Returns:
```
cdolfi, line 175: Should these spaces be here?

8Knot/db_manager/augur_manager.py
```diff
@@ -255,47 +269,38 @@ def is_org(self, org):
...
+
+            # If no environment default, return a generic first option without loading database
+            # This will be properly resolved when user actually searches or selects
+            return {"label": "org: chaoss", "value": "chaoss"}
+
         except Exception as e:
             logging.error(f"Error in initial_multiselect_option: {str(e)}")
             # Return a fallback option in case of any error
-            return {"label": "repo: Fallback Option", "value": -1}
+            return {"label": "org: chaoss", "value": "chaoss"}
```
cdolfi, line 303: This should probably be a generic option as there could be augur instances without the chaoss org. i think that was the reason for the -1 value

8Knot/cache_manager/cache_facade.py
```diff
@@ -123,6 +123,12 @@ def get_uncached(func_name: str, repolist: list[int]) -> list[int]:  # or None
 
     Returns a list of repos that AREN'T resident in cache.
     """
+    # STARTUP OPTIMIZATION: Handle None or empty repolist
+    # This prevents crashes when visualization callbacks trigger with no selected repos
+    # (which happens at startup when repo-choices is None/empty)
```
cdolfi, line 128: i think 126-128 could be simplified to "# Crash prevention at start up when repo-choices is None/empty"

8Knot/cache_manager/cache_facade.py
```diff
@@ -166,9 +172,17 @@ def caching_wrapper(func_name: str, query: str, repolist: list[int], n_repolist_
...
         Exception: If a step fails, will print exception and re-raise.
 
-    Returns:
+        Returns:
         _type_: None
     """
+
+    # STARTUP OPTIMIZATION: Prevent unnecessary database queries with empty repo lists
+    # This avoids expensive query operations when no repositories are selected,
+    # which commonly occurs during app startup before users make selections
```
cdolfi, line 181: i think 179-181 to just 179

8Knot/cache_manager/cache_facade.py
```diff
@@ -207,10 +221,16 @@ def retrieve_from_cache(
     For a given table in cache, get all results
     that having a matching repo_id.
 
-    Results are retrieved by a DataFrame, so column names
+        Results are retrieved by a DataFrame, so column names
```
cdolfi, line 224: Should this spacing be here?

8Knot/cache_manager/cache_facade.py
```diff
@@ -207,10 +221,16 @@ def retrieve_from_cache(
...
     may need to be overridden by calling function.
     """
 
+    # STARTUP OPTIMIZATION: Return empty DataFrame for None/empty repo lists
```
cdolfi, line 228: Can be simplified to "STARTUP OPTIMIZATION: graceful handling of visualization callbacks for None/empty repo lists

MoralCode:
whats the impact of the loading time costs to users when they go to search for something?

It seems like the pre-querying may have been added to make the UI feel more responsive after startup by frontloading the processing. Is there a way we can have the best of both by making this pre-loading asynchronous? i.e:
- fetch the minimum possible data in a blocking manner at startup
- kick of an async background task to prefetch some or all of the data before users need it
- when users search, check if we have the data already before kicking off a possibly long augur DB call

8Knot/pages/index/index_layout.py
```diff
@@ -253,7 +253,7 @@
                             variant="filled",
                             debounce=100,  # debounce time for the search input, since we're implementing client-side caching, we can use a faster debounce
                             data=[augur.initial_multiselect_option()],
-                            value=[augur.initial_multiselect_option()["value"]],
+                            value=[augur.initial_multiselect_option()["value"]],  # Pre-select chaoss org at startup
```
cdolfi, line 256: Should be unique to whatever the user has as default load at start up

docker-compose.yml
```diff
@@ -56,6 +56,7 @@ services:
     env_file:
       - .env
     environment:
+      - DEFAULT_SEARCHBAR_LABEL=chaoss  # Default org to show in searchbar at startup
```
cdolfi, line 59: Should be based off of the value in the .env

8Knot/pages/index/index_callbacks.py
```diff
@@ -597,16 +599,29 @@ def wait_queries(job_ids):
...
     """
 
+    # Don't run queries if no repos selected (None or empty list)
+    if not repos:
+        return []
+
+    # STARTUP OPTIMIZATION: With proper callback chain, we only get here when:
+    # 1. Auto-load callback triggered for default chaoss selection, OR
+    # 2. Search button was clicked for user selections
+    # This maintains fast startup while providing data when needed
```
cdolfi, line 621: remove line 621

8Knot/pages/index/index_callbacks.py
```diff
@@ -428,8 +428,10 @@ def dynamic_multiselect_options(user_in: str, selections, cached_options):
 @callback(
     [Output("results-output-container", "children"), Output("repo-choices", "data")],
     [
-        Input("search", "n_clicks"),
-        State("projects", "value"),
+        Input("search", "n_clicks"),  # Changed back to search button trigger
```
cdolfi, line 431: comments not needed

8Knot/pages/index/index_callbacks.py
```diff
@@ -597,16 +599,29 @@ def wait_queries(job_ids):
...
     Executes queries defined in /queries against Augur
     instance for input Repos; caches results in Postgres.
 
     Args:
-        repos ([int]): repositories we collect data for.
+        repos ([int] | None): repositories we collect data for.
+        n_clicks (int | None): number of times search button was clicked
     """
 
+    # Don't run queries if no repos selected (None or empty list)
```
cdolfi, line 614: "prevents queries if no repos are selected (None or empty list)"

8Knot/db_manager/augur_manager.py
```diff
@@ -255,47 +269,38 @@ def is_org(self, org):
...
+                default_label = os.getenv("DEFAULT_SEARCHBAR_LABEL")
+                logging.warning(
+                    f"INITIAL SEARCHBAR OPTION: Using environment default '{default_label}' without DB lookup"
+                )
+                # Return the environment-specified option directly (assumes it's an org name like "chaoss")
+                return {"label": f"org: {default_label}", "value": default_label.lower()}
+
+            # If no environment default, return a generic first option without loading database
+            # This will be properly resolved when user actually searches or selects
+            return {"label": "org: chaoss", "value": "chaoss"}
```
cdolfi, line 298: needs to be generalized as chaoss org might not be in every db

8Knot/db_manager/augur_manager.py
```diff
@@ -255,47 +269,38 @@ def is_org(self, org):
         Returns:
             bool: whether org name is in orgs
         """
+        # STARTUP OPTIMIZATION: Lazy loading pattern for organization names validation
```
cdolfi, line 272: Instead of separate comment add condensed detail in function description. something like "On demand loading of organization information"

8Knot/db_manager/augur_manager.py
```diff
@@ -244,6 +251,13 @@ def org_to_repos(self, org):
         Returns:
             [int] | None: repo_ids or None
         """
+        # STARTUP OPTIMIZATION: Lazy loading pattern for organization to repositories mapping
```
cdolfi, line 254: Instead of separate comment add condensed detail in function description. something like "On demand loading of list of repos in an org"

8Knot/db_manager/augur_manager.py
```diff
@@ -233,6 +233,13 @@ def repo_id_to_git(self, id):
         Returns:
             git (str): URL of repo
         """
+        # STARTUP OPTIMIZATION: Lazy loading pattern for repo ID to Git URL mappings
```
cdolfi, line 236: Same comment as below

cdolfi (changes requested):
Updated requested for comments, will test more when replication lag is down

8Knot/cache_manager/cache_facade.py
MoralCode: this whole thing is a comment so its just a matter of coding style choice. it seems like if we want it to align with "raises" it shouldnt be, but thats just my opinion

8Knot/db_manager/augur_manager.py
MoralCode: AFAIK the default search option is an environment variable/is configurable so i think this is like a second fallback option and we should probably leave it as is

cdolfi:
Running this branch locally I am not seeing any difference positively in start time, I dont think its making it worse but this doesnt seem to be the source of the start up delay. There is not a replication lag or heavy queries on the DB, so I think there might need further investigation in the actual reason for the variable start up

EngCaioFonseca:
closed due to the new improvements done on 8knot start up and hot reloads PR, which is making it possible to build it significantly faster.

EngCaioFonseca:
closed due to the new improvements done on 8knot start up and hot reloads PR, which is making it possible to build it significantly faster.

PR #873, opened 2025-08-13
Language and File Percentage Calculation Error

sgoggins:
This PR corrects two problems: 

#### Problem 1: Percentage Recalculation After Grouping
```python
# This happens AFTER grouping into "Other"
df_lang["Code %"] = (df_lang["code_lines"] / df_lang["code_lines"].sum()) * 100
df_lang["Files %"] = (df_lang["files"] / df_lang["files"].sum()) * 100
```

When small languages are grouped into "Other", their individual rows are replaced with a single "Other" row. The subsequent percentage calculation uses this modified dataset, leading to incorrect proportions.

#### Problem 2: Data Modification for SVG Files
```python
df.loc[df["programming_language"] == "SVG", "code_lines"] = df["files"]
```
This line modifies the actual line counts for SVG files, which could affect totals if SVG (or misidentified XML as SVG) is present in the data.

### DIFF
--- a/8Knot/pages/index/index_layout.py
+++ b/8Knot/pages/index/index_layout.py
@@ -22,7 +22,7 @@ if os.getenv("AUGUR_LOGIN_ENABLED", "False") != "True":
                         "If you need to collect data on new repositories, please ",
                         html.A(
                             "create a repository collection request",
-                            href="https://github.com/oss-aspen/8Knot/issues/new?template=augur_load.md",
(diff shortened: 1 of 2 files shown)


sgoggins:
I know what's broken here. Will fix before morning. Right now I got a kid anxious about her drivers exam tomorrow

sgoggins:
I have the fix up and working on my end and the percentages now closely match what I see with the raw queries.

PR #875, opened 2025-08-14
Pie chart design

cdolfi:
Note: #871 will need to get merged before this, its built on top of that branch 

This PR styles all of the remaining pie charts


8Knot/pages/affiliation/visualizations/commit_domains.py
```diff
@@ -46,18 +64,30 @@
...
                                 dbc.Col(
                                     [
+                                        dbc.Label(
+                                            "Contributions Required:",
+                                            html_for=f"company-contributions-required-{PAGE}-{VIZ_ID}",
+                                            style={
+                                                "fontSize": "12px",
+                                                "fontWeight": "bold",
+                                                "marginBottom": "0.5rem",
+                                            },
```
MoralCode, line 89: are these one-off styles/exceptions for this element or can they be refactored to the stylesheet to apply to other things?

8Knot/pages/affiliation/visualizations/gh_org_affiliation.py
```diff
@@ -23,10 +23,28 @@
...
+                                color="outline-secondary",
+                                size="sm",
+                                className="about-graph-button",
+                            ),
+                            width="auto",
+                        ),
+                    ],
+                    align="center",
+                    justify="between",
+                    className="mb-3",
```
MoralCode, line 47: it seems like we are making the same changes here across several visualizations.

Would it be better (maybe not now but in future) to have the shared elements defined centrally (in a function or superclass) so that they can just be imported by each visualization so the the specific values for the titles and about information can be passed in while keeping styling somewhat centralized (yet overrideable for special cases)

8Knot/pages/affiliation/visualizations/gh_org_affiliation.py
```diff
@@ -46,63 +64,67 @@
...
                 dcc.Loading(
                     dcc.Graph(id=f"{PAGE}-{VIZ_ID}"),
+                    style={"marginBottom": "1rem"},
+                ),
+                html.Hr(  # Divider between graph and controls
+                    style={
+                        "borderColor": "#909090",
+                        "margin": "1.5rem -1.5rem",
+                        "width": "calc(100% + 3rem)",
+                    }
```
MoralCode, line 74: a divider may not be worth moving into a shared visualization base too, but that might be another helpful thing for consistency/reuse

8Knot/pages/chaoss/visualizations/contrib_importance_pie.py
```diff
@@ -124,48 +153,35 @@
...
                                 ),
                                 dbc.Col(
-                                    [
-                                        dbc.Button(
-                                            "About Graph",
-                                            id=f"popover-target-{PAGE}-{VIZ_ID}",
-                                            color="secondary",
-                                            size="sm",
-                                        ),
-                                    ],
```
MoralCode, line 156: im not seeing where this about graph button got moved to within this visualization

MoralCode (commented):
Seems good! had a couple questions about the code  (the missing about button may be something id consider a blocker) but it seems to run fine and be fairly clean so we can always clean up later once thats answered

8Knot/pages/chaoss/visualizations/contrib_importance_pie.py
cdolfi: line 29

PR #880, opened 2025-08-17
Bar chart redesign

cdolfi:


MoralCode (changes requested):
Screenshot_20250820_165719


Did the Contributors by action type graph get skipped here?

MoralCode:
disregard - it just took me longer than it realistically should have to remember the difference between histogram and bar chart :facepalm:

MoralCode (approved):

PR #884, opened 2025-08-18
Histogram redesign + visualization format change + error fix 

cdolfi:
While reforming contrib by action I fixed a known data preprocessing bug. First time contributors was changed to a bar chart to match the format of the rest of the visualizations


EngCaioFonseca:
Lgtm and test went well.

EngCaioFonseca (approved):

PR #889, opened 2025-08-20
Languages Pie Chart: minor change for other to be based on lines

cdolfi:
Line 174 was unnecessary with the current query and changes other to be based on lines instead of files based on @sgoggins use case where it showed that doing it based on # of files caused problems for certain repos


MoralCode:
So this always calculated "other" by lines of code even if "files" is selected in the graph? 


Screenshot_20250822_130601

cdolfi:
@MoralCode it was based on files but found that it caused for major languages it incorrectly get grouped into other #873

MoralCode (approved):
Seems to run okay on my machine!

Issue #905, opened 2025-08-21
`uv` setup is breaking hot reload


MoralCode:
In order for hot reload to work as intended in #799, we need to be able to mount the current version of the files from the local dev machine into the container
```
volumes:
      - ./8Knot:/opt/app-root/src/
```
this was removed in 5e02a08507dcb3261e010d801ddc5e4dff72311b (possibly due to the issues i will describe below)

doing so will replace the listed directory with the one that is in the container already

this replacement will cause `/opt/app-root/src/.venv` in the container to no longer exist (it doesnt exist in the `8Knot` dir of this repo, nor should it).

This will prevent 8knot from starting up. This likely manifests itself as a failure due to a missing `gunicorn` binary which should be in `/opt/app-root/src/.venv/bin`. However, even if you fix that, you eventually run into an error with python itself because uv will symlink a python binary from your system into this bin folder. that symlink will get put in the container, which will then be unable to traverse it due to not having access to that filepath in the container.

To avoid this, we need to separate the location of the dependencies from the source so we can remount the source dir for hot reload while keeping the dependencies the same.

this should mostly just involve a refactoring of the Dockerfile and compose.yml to allow `.venv` in the container to exist at least one level higher than `src`.

PR #908, opened 2025-08-21
Simplify Dockerfile builds (actually install to system) and fix hot reload

MoralCode:
This stops doing the janky workaround of using virtual environments in our prod container and instead actually installs the dependencies directly to the system (seems like theres already a python environment at `/opt/app-root/` so thats conenient)

the `remove uv binary copied in earlier to hopefully save some space` probably doesn't actually save any space since this PR removes the "builder" container.

This fixes hot reload too.

Everything works on my machine

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -46,6 +46,8 @@ services:
       ]
     # ports:
     #   - 8080:8080
+    volumes:
+      - ./8Knot:/opt/app-root/src/:z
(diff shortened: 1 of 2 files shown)


MoralCode:
Fixes #905

cdolfi:
@MoralCode Rebase branch and squash commits and its good to merge

PR #911, opened 2025-08-22
update repolist alert to look less like debugging output

MoralCode:
Before:
Screenshot_20250822_125720

After:
Screenshot_20250822_125606


cdolfi (approved):
lgtm!

PR #922, opened 2025-08-25
slight formating updates for date range

cdolfi:
format updates, not going too far into it as the graphs will be 1 per row coming soon


cdolfi:
@MoralCode they are showing up differently because they have different inputs

cdolfi:
@MoralCode This is to make the formatting of the date range pickers a bit better for demos this week as well as prevent text overlap. This change will not matter much beyond this week as we are changing to one visualization per row which will change the formatting needs for all of the visualization input. If you look at current dev it might be easier to tell the difference

EngCaioFonseca (approved):
Lgtm!

PR #934, opened 2025-09-04
Refactoring welcome page with dash components - removing nested html divs

EngCaioFonseca:
This PR remove html nested divs and replaces with dash components structure, aiming to align with the dash-centric version of the app, and also to satisfy better clean code principles.


EngCaioFonseca:
Closed due to #939

PR #935, opened 2025-09-08
Upgrade dash mantine and bootstrap related components

MoralCode:
This PR updates the mantine and other bootstrap related components to their latest versions.

Overall this was relatively simple - not nearly as bad as i worried it would be. Changes made to allow this to work:

- One prop of `MultiSelect` changed its name
- the whole layout was wrapped in a `dmc.MantineProvider` using the default theme
- the repo IDs returned by the dash callback for the searchbox were typecast to strings so that the `Option` they eventually get put into behind the scenes doesn't throw an error.

Curious to see how well the CI jobs handle this. I've been able to build and run it locally but haven't super in-depth tested things. Also, it seems like we are maybe overriding so many styles that not much changed...

UI-wise the only thing I notice that's obviously different is that the bubbles representing each repo within the search bar are dark theme instead of light theme. 

Related to #845

### DIFF
--- a/8Knot/app.py
+++ b/8Knot/app.py
@@ -149,26 +149,6 @@ import dash_mantine_components as dmc
 app.layout = dmc.MantineProvider(
     layout,
     forceColorScheme="dark",
-    # theme={
-    #     # add your colors
-    #     "colors": {
-    #          # add your colors
-    #         "deepBlue": ["#E9EDFC", "#C1CCF6", "#99ABF0" "..."], # 10 colors
-    #         # or replace default theme color
-    #         "blue": ["#E9EDFC", "#C1CCF6", "#99ABF0" "..."],   # 10 colors
-    #     },
-    #     "shadows": {
-    #         # other shadows (xs, sm, lg) will be merged from default theme
-    #         "md": "1px 1px 3px rgba(0,0,0,.25)",
-    #         "xl": "5px 5px 3px rgba(0,0,0,.25)",
-    #     },
-    #     "headings": {
-    #         "fontFamily": "Roboto, sans-serif",
-    #         "sizes": {
-    #             "h1": {"fontSize": '30px'},
-    #         },
-    #     },
-    # },
 )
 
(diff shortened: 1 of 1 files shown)


cdolfi:
Why did the whole layout need to be wrapped in a dmc.MantineProvider using the default theme?

cdolfi:
Can you condense some of the commits? 17 seems to be a lot for this size of change

cdolfi:
Im getting some errors running locally when trying to use the search bar

MoralCode:
> Why did the whole layout need to be wrapped in a dmc.MantineProvider using the default theme?

the whole layout needed to be wrapped because that was part of the dash mantine migration guide for the 0.12 to 0.14 migration. The default theme was just me doing it in the most basic way since it didnt seem to affect our theming and it seemed sensible to tackle the theming in a later phase.



> Im getting some errors running locally when trying to use the search bar

ah yep i see that too. I looked into this somewhat deeply and sent a message on the dash mantine components discord server to see whats up/what i missed during the upgrade

EngCaioFonseca:
What specific errors are you guys getting related to the search bar? I was able to run it and check the search bar, and in terms of functionality, couldn't see any new error. The loading of the data for big orgs are taking very long time though

cdolfi:
@EngCaioFonseca My comment was before the last round of updated, its working now. Search bar still needs a bit of work but I think this is an improvement, all other updates make sense in a different PR

cdolfi:
Screenshot 2025-09-10 at 10 15 50 AM
This dropdown is also broken, should have an entry selected on initial load

EngCaioFonseca (approved):
Lgtm!

cdolfi:
@MoralCode While the drop down works the tables are no longer populating. With a change like this please make sure to thoroughly check through the entire app to make sure it functions as intended. This applies the same to reviewing @EngCaioFonseca

cdolfi (changes requested):
remove comments and should be good to go

PR #944, opened 2025-09-17
welcome button fix to being stateful

### LINKED ISSUE #886: Upper nav different behavior
when reviewing #871 i noticed that, in the top nav, clicking "welcome" fully reloads the whole page, wheras just clicking "visualizations" will go to the visualizations part but the sidebar seems to stay where it is and not visibly reload

we should investigate why this is and make the behavior consistent

cdolfi:
fixes #886


EngCaioFonseca (approved):
Lgtm and works well.

PR #945, opened 2025-09-17
check for numericness when looking for org vs repo

### LINKED ISSUE #943: Search bar: Org prefix is getting added when it should be repo:
It looks like repo is never getting prefixed only org: 

Image

MoralCode:
remember how i said in #942 that it probably wouldnt be the last time?

yeah this is an attempt to do the same to fix #943 

fixes #943


MoralCode:
marking as draft so i can audit all the isinstance() calls for similar mistakes

MoralCode:
superseded

PR #947, opened 2025-09-18
remove dropdown custom CSS and rely on mantine for dropdown styling

MoralCode:
supersedes #924

### DIFF
--- a/8Knot/assets/color.css
+++ b/8Knot/assets/color.css
@@ -290,31 +290,6 @@ div.dark-card {
 }
 
 /* =============================================================================
-   DROPDOWN COMPONENTS
-============================================================================= */
-
-/* Dark Dropdown Box Styling */
-.dark-dropdown .Select-control,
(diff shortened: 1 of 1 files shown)


8Knot/assets/color.css
```diff
@@ -289,31 +289,6 @@ div.dark-card {
...
-  min-width: 150px !important;
-}
-
-/* Dark Dropdown Placeholder Styling */
-.dark-dropdown .Select-placeholder,
-.dark-dropdown .Select-value-label {
-  color: white !important;
-}
-
-/* Fix for the vertical bar issue */
```
cdolfi, line 311: this should not be removed, this impacts the date range picker

cdolfi (changes requested):
Screenshot 2025-09-18 at 1 28 58 PM

8Knot/assets/color.css
MoralCode: ah right. fixed!

Issue #941, opened 2025-09-16
Listing and hyperlinks to visualizations


cdolfi:
https://www.figma.com/design/e9ylKHtU7YgWyuXBkI9bm0/Augur-8Knot?node-id=1022-85094&m=dev&t=dJ5TemhCBGr4OL5a-1 Side nav bar, list each of the visualizations and have the user be able to click the name of the visualization and it will scroll down to it

PR #948, opened 2025-09-18
small scale: One graph per row and sidebar links for 

MoralCode:
This PR is a preview of the sidebar linking. It applies to just the overview page and
- rearranges the overview page to be 1 graph per row (#899)
- adds IDs to each graph
- adds sidebar links pointing to each ID

In order to be functional, it requires a change to #939 to make the smooth scroll callback more generic and reusable (and able to detect navigation to fragments) - more info in [this comment](https://github.com/oss-aspen/8Knot/issues/941#issuecomment-3308240925)

Making it a draft so we can get some parallel review going to ensure the design/presentation of the links in the sidebar


cdolfi:
@MoralCode If that is an issue, maybe only show the graphs in the drop down when you are on a certain page? So someone would have to be on the affiliation page to then see the names of the graphs listed out below

cdolfi (changes requested):
Looks great! Very concise, squash commits and resolve the precommit failure and this looks good to me

Issue #959, opened 2025-09-25
search bar formatting issues post #958


cdolfi:
#958 was a major improvement to the formatting of the search bar but still not all the way there. 

Longer repos getting cut off: 

<img width="353" height="168" alt="Image" src="https://github.com/user-attachments/assets/80eda089-cfb8-4656-a875-5d9ca810c38c" />

Some weird formatting when you click into the box: 

https://github.com/user-attachments/assets/89548d53-7d58-4d51-88e0-61add3febf59

MoralCode:
ok, i see the cursor thing only affects chrome based browsers

cdolfi:
closed by #966

cdolfi:
@MoralCode yes good catch

MoralCode:
Is this a blocker for deployment?

MoralCode:
<img width="358" height="68" alt="Image" src="https://github.com/user-attachments/assets/a38a003f-6b62-4752-914e-29e5ccc67e17" />

This appeared when reviewing #978 and seems like it may be worth making part of this issue

cdolfi:
<img width="428" height="174" alt="Image" src="https://github.com/user-attachments/assets/092e9726-a221-4077-8eda-53fc7e9d558e" />

Still an issue as of 11/19

PR #965, opened 2025-09-25
Resolve cursor visibility issue on search bar

EngCaioFonseca:
Solves part of #959 

Solves visibility of text cursor, by adjusting positioning and removing overlap.


8Knot/assets/cursor-fix.css
```diff
@@ -0,0 +1,39 @@
...
+    position: static !important;
+    top: auto !important;
+}
+
+
+/* Ensure text cursor is visible with white color */
+
+.mantine-MultiSelect-input {
+    caret-color: #ffffff !important;
+}
```
MoralCode, line 35: IMO this is entirely too much CSS to fix one issue.

I have also seen a lot of very similar CSS in other files

MoralCode:
im going through and doing a lot of CSS cleanup as a result of my incomplete dropdown styling and im cleaning up a whole bunch of stuff. I suspect I can fix this issue too

MoralCode:
This PR is also likely somewhat ineffective because the core search bar styles are being set in inline styles. attempting to override them with proper CSS is going to result in more cursed code/tech debt

EngCaioFonseca:
I'm working on a different PR removing all inline styles, and also placing them in proper files, because most of them are placed under just a few or one css file altogether.

PR #966, opened 2025-09-26
fix Searchbar cursor gligh on chrome

MoralCode:
Solves part of #959 

Solves visibility of text cursor by setting the input box to be transparent, preventing part of the cursor from being covered with a color.

This is mututally exclusive with #965


cdolfi (approved):
lgtm

Issue #915, opened 2025-08-25
Format issue with date filter


cdolfi:
<img width="681" height="794" alt="Image" src="https://github.com/user-attachments/assets/ea0a286b-d30c-4de3-a975-2aba1e576811" />

This issue has to do with the formatting of the inputs, the overlap and overall clunky formatting

MoralCode:
while #967 doesnt completely solve the "date range wrapping on two lines" issue, once #900 is fixed (one graph per row) it should be less likely to impact narrower screens

PR #967, opened 2025-09-26
update width for date range picker so both halves show on one line

MoralCode:
This bumps the width of the date picker (which seems to always be in a row by itself) from 5 columns to 7 (and adds a `dbc.Col` wrapper where necessary to accomplish this). This allows both ends of the date range to be side-by-side.

related to #915

### DIFF
--- a/8Knot/pages/contributions/visualizations/issues_over_time.py
+++ b/8Knot/pages/contributions/visualizations/issues_over_time.py
@@ -100,18 +100,21 @@ gc_issues_over_time = dbc.Card(
                             justify="start",
                         ),
                         dbc.Row(
-                            dcc.DatePickerRange(
-                                id=f"date-picker-range-{PAGE}-{VIZ_ID}",
-                                min_date_allowed=dt.date(2005, 1, 1),
-                                max_date_allowed=dt.date.today(),
-                                initial_visible_month=dt.date(dt.date.today().year, 1, 1),
-                                clearable=True,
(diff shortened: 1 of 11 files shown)


8Knot/pages/contributions/visualizations/issues_over_time.py
```diff
@@ -100,18 +100,21 @@
...
+                                    initial_visible_month=dt.date(dt.date.today().year, 1, 1),
+                                    clearable=True,
+                                    start_date=dt.date(
+                                        dt.date.today().year - 2,
+                                        dt.date.today().month,
+                                        dt.date.today().day,
+                                    ),
+                                    className="dark-date-picker",
                                 ),
-                                className="dark-date-picker",
```
cdolfi, line 114: Can you make this a 1 line change as it is for all of the initial files?

cdolfi (changes requested):
change to make all of the files a 1 line change

8Knot/pages/contributions/visualizations/issues_over_time.py
MoralCode: there were ~2 visualizations where I could not do this with a one line change because they were not already wrapped in a `dbc.Col`. I tried to add a `width` parameter to the existing `dbc.Row` but this caused the app to crash and not start up because `width` is not a valid argument for `Row`. these multiline changes are necessary (all they do is wrap the picker in a Col)

PR #968, opened 2025-09-26
enable integer typecasts to handle nulls

MoralCode:
i noticed that when loading a repo that isnt cached, the ossf scorecard and repo info tables seem to have issues with the `int()` typecast.

I suspect this is because a null value is getting passed in as the set repo (potentially a race condition between the callback that sets the selected repo to the default value and the callbacks processing the graphs). This causes the typecasts to fail loudly.

This PR allows null values to pass through (and hopefully be handled as they would have prior to the mantine update that caused all this


cdolfi (approved):
lgtm!

PR #970, opened 2025-10-01
Removal of prefix

EngCaioFonseca:
Removes prefix from repo names, aiming to cut the long names from search.

### DIFF
--- a/8Knot/pages/index/search_utils.py
+++ b/8Knot/pages/index/search_utils.py
@@ -190,3 +190,36 @@ def token_match_score(tokens: List[str], label: str) -> float:
     if match_scores:
         return sum(match_scores) / len(match_scores)
     return 0.0
+
+
+def clean_repo_name(repo_name: str) -> str:
+    """
+    Clean repository names by removing GitHub URL prefixes.
+
+    Args:
+        repo_name: The original repository name/URL
+
+    Returns:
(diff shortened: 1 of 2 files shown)


8Knot/pages/index/search_utils.py
```diff
@@ -190,3 +190,36 @@ def token_match_score(tokens: List[str], label: str) -> float:
...
+        repo_name: The original repository name/URL
+
+    Returns:
+        Cleaned repository name without URL prefixes
+    """
+    if not repo_name:
+        return repo_name
+
+    # Remove common GitHub URL prefixes
+    prefixes_to_remove = [
```
cdolfi, line 209: Have you seen examples of each of these in the db?

cdolfi (changes requested):
From looking at the db I dont think the list of prefixes is necessary. The only github prefix is "https://github.com" . One thing to keep in mind is if you want to solve for github/gitlab now or in a future pr. Once we have gitlab consistently working "Github Repo:" and "Gitlab Repo:" would be the proper labels

EngCaioFonseca:
Now it accounts for github and gitlab as well

cdolfi (approved):
lgtm! Good structure as well when we have new repo sources in the future

PR #973, opened 2025-10-01
Refactor visualizations - trial PR

MoralCode:
Now that I have discovered a pattern for building custom components (as you may have seen in previous PRs), i have used this new superpower to ~~extract all demons~~ move a good chunk of the shared components of a visualization into one central place for maximum reuse.

This means the files defining each individual visualization are less bloated with stuff that "just needs to be included because every visualization needs it" (such as the dividers, and various structural elements).

This has been applied to one graph on the CHAOSS page if anyone is curious (meant to use the repo general info page but opened the wrong file and only realized after I was done) 

## Summary by CodeRabbit

* **New Features**
  * Introduced a reusable All‑In‑One visualization card with integrated title, About popover, loading state, graph area, and controls.

* **Refactor**
  * Replaced handcrafted chart cards across multiple pages with the new visualization component for consistent layout and simpler wiring.
  * Consolidated title, controls, and info popovers into the unified component.

* **Documentation**
  * Updated guidance and path/name references to reflect the new visualization patterns and project structure.

### DIFF
--- a/8Knot/components/visualization.py
+++ b/8Knot/components/visualization.py
@@ -1,0 +1,102 @@
+from dash import html, dcc, Input, State, Output, callback, MATCH
+import dash_bootstrap_components as dbc
+
+
+# All-in-One Components should be suffixed with 'AIO'
+class VisualizationAIO(dbc.Card):
+    def __init__(
+        self,
+        page: str,
+        viz_id: str,
+        graph_info="",
+        class_name="",
+        controls=None,
+        title: str = "",
+        id: Optional[str] = None,
+    ):
+        """
+        Common visualization shell to be shared by all visualizations
+
+        Args:
+            page (str): The name of the page this visualization is part of
+            viz_id (str): a unique id for this visualization
+            graph_info (str): The description of this graph giving more information on what it describes and where its data came from. Displayed in a popover.
+            class_name (str): Any custom class names to associate with this card
+            controls (list): A list of form elements to display within the lower form Row at the bottom of the graph
+            title (Optional[str]): a static title. If none, the title will be fetched from a callback with the id "graph-title-{page}-{viz_id}". Defaults to none.
+            id (Optional[str]): an identifier to use to jump to the card. Primarily intended for navigation, not for styling. Defaults to none, which causes {page}-{viz_id} to be used
+        """
+        self.page = page
+        self.viz_id = viz_id
+        self.nav_id = id
+
+        if controls is None:
+            controls = []
+
+        # Define the component's layout
+        super().__init__(
+            [
+                dbc.CardBody(
+                    [
+                        dbc.Row(
+                            [
+                                dbc.Col(html.H3(title, id=f"graph-title-{page}-{viz_id}", className="card-title")),
+                                dbc.Col(
+                                    dbc.Button(
+                                        "About Graph",
+                                        id={"type": "popover-target", "index": f"{page}-{viz_id}"},
+                                        color="outline-secondary",
+                                        size="sm",
+                                        className="about-graph-button",
+                                    ),
+                                    width="auto",
+                                ),
+                            ],
+                            align="center",
+                            justify="between",
+                            className="mb-3",
+                        ),
+                        dbc.Popover(
+                            [
+                                dbc.PopoverHeader("Graph Info:"),
+                                dbc.PopoverBody(graph_info),
+                            ],
+                            id={"type": "popover", "index": f"{page}-{viz_id}"},
+                            target={"type": "popover-target", "index": f"{page}-{viz_id}"},
+                            placement="top",
+                            is_open=False,
+                        ),
+                        dcc.Loading(
+                            dcc.Graph(id=f"{page}-{viz_id}"),
+                            style={"marginBottom": "1rem"},
+                        ),
+                        html.Hr(className="card-split") if controls else None,  # Divider between graph and controls
+                        dbc.Form(
+                            [
+                                dbc.Row(
+                                    controls,
+                                    align="center",
+                                    justify="start",
+                                ),
+                            ]
+                        )
+                        if controls
+                        else None,
+                    ],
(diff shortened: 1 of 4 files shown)


cdolfi (changes requested):
Like the idea of this! Ill be back to review when I have the proper brain power. I have strong opinions about the visualization structure so marking this as "request changes" to block merging before I get a chance to give a thorough review

8Knot/components/visualization.py
```diff
@@ -0,0 +1,87 @@
...
+class VisualizationAIO(dbc.Card):
+    def __init__(self, page: str, viz_id: str, graph_info="", class_name="", controls=None):
+        """
+        Common visualization shell to be shared by all visualizations
+
+        Args:
+            page (str): The name of the page this visualization is part of
+            viz_id (str): a unique id for this visualization
+            graph_info (str): The description of this graph giving more information on what it describes and where its data came from. Displayed in a popover.
+            class_name (str): Any custom class names to associate with this card
```
cdolfi, line 24: any reason why this cant be based on a combo of page and viz_id?

8Knot/pages/chaoss/visualizations/contrib_importance_pie.py
```diff
@@ -17,169 +17,111 @@
...
+                initial_visible_month=dt.date(dt.date.today().year, 1, 1),
+                clearable=True,
+                className="dark-date-picker",
+            ),
+            width=7,
         ),
     ],
-    className="dark-card",
+    class_name="dark-card",
     id="bus-factor",
```
cdolfi, line 119: Is this needed?

cdolfi (changes requested):
a few small questions but overall a fan!

cdolfi:
The first PR id want to merge of this would actually be an update to the visualization template and instructions. The an overhaul of the visualizations themselves

8Knot/pages/chaoss/visualizations/contrib_importance_pie.py
MoralCode: Unlikely, unless it got renamed in the template class. I can look tomorrow

8Knot/components/visualization.py
MoralCode: What do you mean? Like set the class name to a combination of the page and viz?

cdolfi: thinking about it more not sure when there would ever be a need for a custom class name in this setting

8Knot/components/visualization.py
```diff
@@ -0,0 +1,87 @@
...
+            [
+                dbc.CardBody(
+                    [
+                        dbc.Row(
+                            [
+                                dbc.Col(html.H3(id=f"graph-title-{page}-{viz_id}", className="card-title")),
+                                dbc.Col(
+                                    dbc.Button(
+                                        "About Graph",
+                                        id={"type": "popover-target", "index": f"{page}-{viz_id}"},
```
cdolfi, line 47: Can you explain this change in syntax a bit?

8Knot/components/visualization.py
MoralCode: Yeah, ideally they'd all be handled by the same common classes, this would likely just be a pass through in case theres a special case in the future where we want to do something special on one visualization card.

8Knot/components/visualization.py
MoralCode: You mean using an object as the ID? That part of a feature of dash called pattern matching callbacks https://dash.plotly.com/pattern-matching-callbacks

It basically means we dont have to write an individual callback for each and every "about this graph" popover, and can instead have one callback handle all of them

cdolfi: Nice, I think this is new from when we first implemented this. Good change

8Knot/components/visualization.py
MoralCode: So in terms of how this is actually used, this would be for passing through like the `dark-card` class. These get added to the existing classes. page and viz id are also passed in so this viz component can also construct the class names in the format that we expect them to be. This passthrough is more for like applying styling classes, especially if we want to use some kind of styling thats part of some "not controlled by us" CSS (i.e. bootstrap)

8Knot/pages/chaoss/visualizations/contrib_importance_pie.py
MoralCode: I suspect this was just from me renaming the parameter in the viz component to accept the suggestion from the linter in my editor that names should be snake case (per python convention).

Do we have a different project level convention? happy to adjust the names and linter settings to match that if we do

cdolfi:
@MoralCode template change looks good! I think there is a few more updates needed from the overhaul in general but I am okay with those improvements made in the next visualization that is made (hoping to do the fedora ones in dec). https://github.com/oss-aspen/8Knot/blob/dev/docs/new_viz_guidance.md do you think this is still representative?

cdolfi:
@MoralCode Can you do the one spell fix and squash commits if needed?

8Knot/components/visualization.py
```diff
@@ -0,0 +1,102 @@
...
+            id=(self.nav_id if self.nav_id is not None else f"{self.page}-{self.viz_id}"),
+        )
+
+    # callback for graph info popover
+    @callback(
+        Output({"type": "popover", "index": MATCH}, "is_open"),
+        [Input({"type": "popover-target", "index": MATCH}, "n_clicks")],
+        [State({"type": "popover", "index": MATCH}, "is_open")],
+    )
+    def toggle_popover(n, is_open):
```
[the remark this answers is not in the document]
MoralCode: the docs examples dont do this

https://dash.plotly.com/all-in-one-components#all-in-one-component-limitations

cdolfi (approved):
lgtm!

PR #980, opened 2025-10-07
removal of duplicates/similar CSS. Prioritizing global variables.

### LINKED ISSUE #960: Welcome page tabs considerations and improvements
- [ ] Make the arrows white/consider a different way of conveying the information 
- [ ] Consider making the text for the titles of the subsection a blue shade
- [ ] work on the tab colors

EngCaioFonseca:
Addressing #960 

CSS:
- Removing duplicates/similar variables. 
- Prioritizing global variables.
- Removing unused variables. 
- Added specificity while using the global variables;
- removal of multiple !important commands.

## Summary by CodeRabbit

* **Style**
  * Migrated the app to a unified global token system for colors, spacing, typography, radii, shadows and transitions.
  * Added a scoped landing-page namespace and tokenized hero, logo, title, subtitle, CTA, tabs, cards and related interactive states.
  * Standardized sidebar, navigation, search/dropdown, input placeholder and Mantine-related styling to token-driven values.
  * Updated responsive rules and typography/blockquote/data-display styles for more consistent theming across breakpoints.

### DIFF
--- a/8Knot/assets/color.css
+++ b/8Knot/assets/color.css
@@ -1,412 +1,70 @@
-/* ==== CSS VARIABLES FOR CONSISTENT THEMING ==== */
+/* ==================================================================
+   COLOR VARIABLES ONLY
+
+   This file contains ONLY color-related CSS variables.
+   Layout tokens (spacing, border-radius, transitions) are in main_layout.css
+   All component styling has been moved to main_layout.css
+   ================================================================== */
+
 :root {
-  --placeholder-bg-color: #1D1D1D;
-  --placeholder-text-color: #9c9c9c;
-  --placeholder-font-size: 16px;
-  --placeholder-line-height: 20px;
-  --placeholder-top-offset: 1px;
-  --multiselect-hover-bg: #404040;
-  --color-primary: #F0F9FF;
-  --color-secondary: #b5b683;
-  --color-accent: #BAE6FD;
-  --color-muted: #aaa;
-  --color-white: #ffffff;
-  --color-black: #000000;
-  --color-dark-bg: #1D1D1D;
-  --color-card-bg: #272b30;
-  --color-border: #292929;
-  --color-topbar-bg: #242424;
-  --color-text-light: #f0f0f0;
-  --spacing-sm: 0.25rem;
-  --spacing-md: 0.5rem;
-  --spacing-lg: 1rem;
-  --border-radius: 1rem;
-  --baby-blue-100: #DFF0FB;
-  --baby-blue-200: #A8D9F5;
-  --baby-blue-300: #76C5EF;
-  --baby-blue-400: #3FB0E9;
-  --baby-blue-500: #199AD6;
-  --baby-blue-600: #147AAE;
-  --baby-blue-700: #0F5880;
-  --content-subtle: #C7C7C7;
+    /* Base colors */
+    --color-primary: #F0F9FF;
+    --color-secondary: #b5b683;
+    --color-accent: #BAE6FD;
+    --color-muted: #aaa;
+    --color-white: #ffffff;
+    --color-black: #000000;
+    /* Background colors */
+    --color-dark-bg: #1D1D1D;
+    --color-card-bg: #272b30;
+    --color-border: #292929;
+    --color-topbar-bg: #242424;
+    --bg-primary: #1D1D1D;
+    --bg-secondary: #242424;
+    --bg-tertiary: #292929;
+    /* Text colors */
+    --color-text-light: #f0f0f0;
+    --color-text-secondary: #9c9c9c;
+    --color-text-muted: #CCCCCC;
+    /* Border colors */
+    --border-color-ui: #404040;
+    --border-color-dark: #444;
+    --border-color-light: #666666;
+    /* Interactive state colors */
+    --hover-bg: #404040;
+    --hover-border: #606060;
+    --multiselect-hover-bg: #404040;
+    /* Baby blue palette */
+    --baby-blue-100: #DFF0FB;
+    --baby-blue-200: #A8D9F5;
+    --baby-blue-300: #76C5EF;
+    --baby-blue-400: #3FB0E9;
+    --baby-blue-500: #199AD6;
+    --baby-blue-600: #147AAE;
+    --baby-blue-700: #0F5880;
+    /* Accent colors */
+    --accent-blue: #119DFF;
+    --accent-teal: #008489;
+    --content-subtle: #C7C7C7;
+    /* Gray scale */
+    --gray-dark: #505050;
+    --gray-medium: #555555;
+    --gray-light: #787878;
+    --gray-lighter: #E5E5E5;
+    --border-divider: #909090;
+    /* Component-specific colors */
+    --placeholder-bg-color: #1D1D1D;
+    --placeholder-text-color: #9c9c9c;
+    --button-light-bg: #f8f9fa;
+    --button-text-light: #fff9f9;
+    /* Table colors */
+    --table-bg-dark: #222;
+    --table-row-hover: #2e2e2e;
+    --table-row-dark: rgb(33, 33, 33);
+    /* Shadow colors */
+    --shadow-color-sm: rgba(0, 0, 0, 0.1);
+    --shadow-color-md: rgba(0, 0, 0, 0.15);
+    --shadow-color-lg: rgba(0, 0, 0, 0.2);
+    --shadow-focus: rgba(64, 64, 64, 0.25);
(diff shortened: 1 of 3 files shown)


8Knot/assets/landing_page.css
```diff
@@ -49,18 +29,18 @@
...
     text-align: center;
-    padding: var(--landing-spacing-huge) var(--landing-spacing-lg) var(--landing-spacing-xxl) var(--landing-spacing-lg);
-    color: var(--landing-text-primary);
+    padding: 60px var(--spacing-lg) 40px var(--spacing-lg);
+    color: var(--text-primary);
 }
 
 .landing-logo-section {
-    margin-bottom: var(--landing-spacing-xxl);
+    margin-bottom: 40px;
```
MoralCode: it seems like we are reusing this 40px value a couple times in this file, so maybe it does deserve its own variable

8Knot/assets/landing_page.css
```diff
@@ -172,17 +152,17 @@
...
-.card-container {
-    background: var(--color-card-bg) !important;
-    border: 1px solid var(--color-border) !important;
-    border-radius: var(--border-radius-lg) !important;
-    box-shadow: 0 4px 6px rgba(0, 0, 0, 0.1) !important;
+.landing-page .card-container {
+    background: var(--color-card-bg);
+    border: 1px solid var(--color-border);
+    border-radius: var(--border-radius-lg);
+    box-shadow: 0 4px 6px rgba(0, 0, 0, 0.1);
```
MoralCode: nice change! love the use of more specific selectors instead of `!important`

MoralCode (commented):
Overall seems okay!

I generally am not a fan of replacing variables with magic/unexplained values, especially if those values are used in multiple places. I think we should keep some variables for those cases where we are reusing the same values a lot for the same purposes.

Havent actually run this yet so holding off on full approval, but i think this is a great reduction in the number of duplicate variables and I love that the `!important` flags are being replaced with more specific/appropriate selectors

8Knot/assets/landing_page.css
cdolfi: Agree with this take as well

8Knot/assets/landing_page.css
```diff
@@ -69,67 +49,67 @@
...
+    color: var(--text-primary);
+    margin-bottom: var(--spacing-md);
     line-height: 1.2;
     margin-top: 0;
 }
 
 .landing-subtitle {
-    font-size: var(--landing-spacing-md);
-    color: var(--landing-text-secondary);
+    font-size: 16px;
```
cdolfi: Is this a repeated font size?

cdolfi: See it below, seems like there is a need for some font size variable, probably where the rest of the global variables are

8Knot/assets/landing_page.css
```diff
@@ -69,67 +49,67 @@
...
+    color: var(--text-secondary);
     margin-top: 0;
 }
 
 .landing-learn-button {
-    background-color: var(--landing-button-bg) !important;
-    color: var(--landing-button-text) !important;
-    padding: var(--landing-spacing-sm) var(--landing-spacing-xl);
-    border-radius: var(--landing-border-radius);
+    background-color: #f8f9fa;
```
cdolfi: Are these variables reused anywhere?

cdolfi (changes requested):
nice overall looks good, just seems like there needs to be around of making some variables for reused values and it will be good to go

cdolfi (changes requested):
I agree with the idea of using the variables on the main css styling sheets. Most of the colors that are now hex values are used multiple times and most are already defined in color.css . Change to variable names and add to the color.css

8Knot/assets/color.css
```diff
@@ -1,412 +1,578 @@
...
-  margin-bottom: 0;
-  margin-right: 0.5rem;
-  margin-bottom: 0;
+    font-size: 14px;
+    font-weight: 500;
+    color: var(--color-white);
+    margin-bottom: 0;
+    margin-right: 0.5rem;
+    margin-bottom: 0;
 }
```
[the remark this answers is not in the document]
EngCaioFonseca: Resolved

EngCaioFonseca:
Tried to not add the color.css with all the formatting changes, but once I started the changes in there, there was no way around. The alternative would be prob more time spent in not having in this one PR as well.

8Knot/assets/landing_page.css
```diff
@@ -191,253 +191,246 @@
...
-        padding: 6px 12px !important;
-        min-width: 120px !important;
+    .landing-page .tabs-header {
+        padding: 0px var(--spacing-sm);
+    }
+    .landing-page .tabs-header .nav-tabs .nav-item:nth-child(1) .nav-link,
+    .landing-page .tabs-header .nav-tabs .nav-item:nth-child(2) .nav-link,
+    .landing-page .tabs-header .nav-tabs .nav-item:nth-child(3) .nav-link,
+    .landing-page .tabs-header .nav-tabs .nav-item:nth-child(4) .nav-link,
+    .landing-page .tabs-header .nav-tabs .nav-item:nth-child(5) .nav-link {
```
MoralCode, line 390: note to self, im pretty sure tabs-header is only used on the landing page, so this is probably okay, but if we wanted to reuse it elsewhere it may be a problem

8Knot/assets/landing_page.css
```diff
@@ -581,248 +574,68 @@
...
+    border: none;
+    border-radius: var(--border-radius-md);
+    box-shadow: none;
 }
 
-.sidebar-navigation .card-body {
-    background: transparent !important;
+.landing-page .sidebar-navigation .card-body {
+    background: transparent;
 }
```
MoralCode, line 609: are we sure this (or any other places where the `.landing-page` class was added) wont cause those styles to break in places that `.sidebar-navigation` is used without being part of a `.landing-page`?

8Knot/assets/main_layout.css
```diff
@@ -6,31 +6,9 @@
...
-    --accent-success: #0F5880;
-    --spacing-xs: 0.25rem;
-    --spacing-sm: 0.5rem;
-    --spacing-md: 1rem;
-    --spacing-lg: 1.5rem;
-    --spacing-xl: 2rem;
-    --border-radius: 8px;
-    --border-radius-lg: 12px;
-    --border-radius-xl: 20px;
-    --transition: all 0.3s ease;
```
MoralCode, line 32: these variables dont represent color - do we want them to belong in color.css anyway?

MoralCode: theyre all globals so it probably doesnt matter, but it may be nice to organize things and stay consistent (or ideally reduce the CSS to being simple enough that it can all be in one file)

MoralCode (commented):
Just scrolled through the full diff and had some minor comments/questions. 

I should probably run this soon to make sure the styles look okay, but overall im in favor of this since the diff does overall reduce the amount of CSS being used.

I think it would be a great goal (not neceserily for this PR though) to see if we can  simplify or rely on libraries enough to allow us to get the 8knot CSS to be small and understandable enough that it fits in one file.

8Knot/assets/main_layout.css
EngCaioFonseca: we can maybe rename the title of the file eventually, to instead of color.css have something like global_val.css that is if we want everything together in one file.

8Knot/assets/landing_page.css
EngCaioFonseca: they wouldn't because the structures are different, in landing_page this is using dbc.cards, and the main doesn't use them, but to avoid confusion, I'm actually deleting it, as in my review here, we're not using it anymore. I also tested the app, and the sidebar works fine after removing it.

cdolfi:
Color in color, and layout in main

8Knot/assets/color.css
```diff
@@ -1,412 +1,96 @@
...
+   - NO component-specific colors (those belong in component files)
+   - NO spacing, typography, border radius, or other design tokens
+   - NO component styling
+
+   Component-specific colors belong in:
+   - main_layout.css (searchbar, tables, multiselect, etc.)
+   - landing_page.css (buttons, tabs, features, etc.)
+   ================================================================== */
 
-/* ==== FORM COMPONENTS ==== */
```
cdolfi, line 120: Are all of these deletes correct?

cdolfi (changes requested):
I ran this locally and some of the needed styling was lost, 

example: 
Screenshot 2025-11-12 at 4 49 45 PM

cdolfi (approved):
lgtm

PR #992, opened 2025-10-29
feat: Add auto-trigger search on dropdown selection

EngCaioFonseca:
PR enables the user to trigger the querying and loading when selecting org/repo also straight from the dropdown

### DIFF
--- a/8Knot/pages/index/index_callbacks.py
+++ b/8Knot/pages/index/index_callbacks.py
@@ -433,11 +433,12 @@ def dynamic_multiselect_options(user_in: str, selections, cached_options):
 
 
 # callback for repo selections to feed into visualization call backs
+# Modified to trigger on EITHER dropdown selection OR search button click
 @callback(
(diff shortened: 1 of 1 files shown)


MoralCode (changes requested):
I really dont think this is going to help our performance or user experience. ESPECIALLY until we fix the bug that causes queries started in a previous tab to not get cancelled when the user loads more data (i.e. moves to a new tab). Without that fix (or reorganizing the cache to be per-repo), this change will just stack more and more and more unnecessary queries on top of each other

For user experience hotpatches, I think we can just add a little popup warning when the searchbox content doesnt match the actually selected repos list and the user is on a page where they are viewing graphs (to prompt them to hit search, shake the search icon, etc)


for performance, maybe we can find ways to optimize our queries so that we pull less data from augur. at least by default. then the data is likely to load faster. Maybe we can test locally with pages that only have one graph on them.

PR #1007, opened 2025-11-13
Refactor index_layout: Extract nested components into modular utility functions

EngCaioFonseca:
Cleaning/refactoring code. Removing nesting, creating functions into their own utils files. 

* Search bar nesting removed. Functions moved into search_utils;
* Login banner and warning functions placed into new file login_utils.

## Summary by CodeRabbit

* **Style**
  * Enhanced search bar and multiselect component styling with improved layout and visual hierarchy.

* **New Features**
  * Added login disabled banner that displays informational messaging when authentication is unavailable.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/8Knot/pages/index/login_utils.py
+++ b/8Knot/pages/index/login_utils.py
@@ -1,0 +1,216 @@
+"""
+Login utilities
+Provides modular components for login banner and navigation elements.
+
+Generic Builders (All-in-One Pattern):
+    create_nav_item: Build any type of navigation item
+
+Login-Specific Functions:
+    create_login_navbar: Main login navigation bar
+    create_login_disabled_banner: Banner shown when login is disabled
+    is_login_enabled: Check login status from environment
+"""
+import os
+from typing import Optional, Dict, Any, Union, List, TYPE_CHECKING
+from dash import html, dcc
+import dash_bootstrap_components as dbc
+
+# Use TYPE_CHECKING to avoid circular imports
+if TYPE_CHECKING:
+    from db_manager.augur_manager import AugurManager
+
+
+# ===== GENERIC COMPONENT BUILDERS (All-in-One Pattern) =====
+
+
+def create_nav_item(
+    item_id: str,
+    content: str,
+    disabled: bool = False,
+    href: Optional[str] = None,
+    external_link: bool = False,
+    target: Optional[str] = None,
+) -> dbc.NavItem:
+    """
+    Generic navigation item builder following Dash All-in-One Components pattern.
+
+    Args:
+        item_id: Unique identifier for the nav item
+        content: Text content for the nav link
+        disabled: Whether the nav item is disabled
+        href: Link URL (optional)
+        external_link: Whether link is external
+        target: Link target attribute
+
+    Returns:
+        dbc.NavItem containing a configured NavLink
+    """
+    return dbc.NavItem(
+        dbc.NavLink(
+            content,
+            id=item_id,
+            disabled=disabled,
+            href=href,
+            external_link=external_link,
+            target=target,
+        ),
+    )
+
+
+# ===== LOGIN COMPONENT FUNCTIONS =====
+
+
+def create_login_disabled_alert():
+    """
+    Create the alert content for the login disabled banner.
+    Uses CSS classes from main_layout.css for consistent styling.
+
+    Returns:
+        dbc.Alert component with login disabled message
+    """
+    return dbc.Alert(
+        [
+            html.H4(
+                "Login is Currently Disabled",
+                className="alert-heading login-banner-heading",
+            ),
+            html.P(
+                [
+                    "If you need to collect data on new repositories, please ",
+                    html.A(
+                        "create a repository collection request",
+                        href="https://github.com/oss-aspen/8Knot/issues/new?template=augur_load.md",
+                        target="_blank",
+                        className="login-banner-link",
+                    ),
+                    ".",
+                ],
+                className="login-banner-text",
+            ),
+        ],
+        color="light",
+        dismissable=True,
+        duration=60000,
+        id="login-disabled-banner",
+        className="mb-0 login-banner-alert",
+    )
+
+
+def create_login_disabled_banner():
+    """
+    Create the banner displayed when login is disabled.
+    Uses CSS class .login-banner-container for positioning.
+
+    Returns:
+        html.Div containing the login disabled banner, or None if login is enabled
+    """
+    if not is_login_enabled():
+        return html.Div(
+            create_login_disabled_alert(),
+            className="login-banner-container",
+        )
+    return None
+
+
+def create_login_container():
+    """
+    Create the login container with loading indicator.
+
+    Returns:
+        dcc.Loading component wrapping the nav-login-container
+    """
+    return dcc.Loading(
+        children=[
+            html.Div(
+                id="nav-login-container",
+                children=[],
+            ),
+        ]
+    )
+
+
+def create_login_popover():
+    """
+    Create the login failed popover.
+
+    Returns:
+        dbc.Popover component for login failure notifications
+    """
+    return dbc.Popover(
+        children="Login Failed",
+        body=True,
+        id="login-popover",
+        is_open=False,
+        placement="bottom-end",
+        target="nav-dropdown",
+    )
+
+
+def create_login_nav(augur_manager: "AugurManager"):
+    """
+    Create the login navigation component with all items.
+    Uses the generic create_nav_item builder directly.
+
+    Args:
+        augur_manager: The AugurManager instance with user_account_endpoint attribute
+
+    Returns:
+        dbc.Nav component with all login navigation items
+    """
+    return dbc.Nav(
+        [
+            create_login_container(),
+            create_nav_item(
+                item_id="refresh-button",
+                content="Refresh Groups",
+                disabled=True,
+            ),
+            create_nav_item(
+                item_id="manage-group-button",
+                content="Manage Groups",
+                disabled=True,
+                href=f"{augur_manager.user_account_endpoint}?section=tracker",
+                external_link=True,
+                target="_blank",
+            ),
+            create_nav_item(
+                item_id="logout-button",
+                content="Log out",
+                disabled=True,
+                href="/logout/",
+                external_link=True,
+            ),
+            create_login_popover(),
+        ]
+    )
+
+
+def create_login_navbar(augur_manager: "AugurManager"):
+    """
+    Create the login navbar based on AUGUR_LOGIN_ENABLED environment variable.
+
+    Args:
+        augur_manager: The AugurManager instance (only used if login is enabled)
+
+    Returns:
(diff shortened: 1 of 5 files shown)


8Knot/pages/index/login_utils.py
```diff
@@ -0,0 +1,216 @@
+"""
+Login utilities
+Provides modular components for login banner and navigation elements.
+"""
+import os
+from dash import html, dcc
+import dash_bootstrap_components as dbc
+
+
+def create_login_disabled_alert():
```
MoralCode, line 63: rather than functions called "create <thing>", i think we should consider using dash all in one components: dash.plotly.com/all-in-one-components like i did for the visualization refactor to create truly generic components.

For example for this function, i would make an Alert component that allowed the title, message, and colors to be customized, so we can reuse it for many types of alerts across the whole site.

The core point of functions is to make code reusable by allowing you to encapsulate a small unit of functionality and expose certain parameters to change its operation. If we are simply moving things into a function with no args without making use of these arguments to allow this function to be reused, then we are just creating a situation where future-us is going to just create a function called `create_repository_error_alert()` or `my_custom_alert()` and we are going to be back to having a problem of duplicate or otherwise hard to maintain code.

All in one components are, i think, a better option because they are in their own separate folder and thus almost have to (well not technically but i guess in as close as you can get to "socially" with code) be made generically because they are farther away in the code from where they are used, and thus could be imported by anything else in the codebase.

8Knot/pages/index/login_utils.py
```diff
@@ -0,0 +1,216 @@
...
+        className="mb-0",
+        style={
+            "backgroundColor": "#DFF0FB",
+            "borderColor": "#0F5880",
+            "border": "1px solid #0F5880",
+            "borderLeft": "5px solid #0F5880",
+            "boxShadow": "0 2px 8px rgba(0, 0, 0, 0.15)",
+            "maxWidth": "400px",
+            "padding": "15px",
+            "zIndex": "1000",
```
MoralCode: why are these styles inline?

8Knot/pages/index/login_utils.py
```diff
@@ -0,0 +1,216 @@
...
+        dbc.NavItem with refresh button
+    """
+    return dbc.NavItem(
+        dbc.NavLink("Refresh Groups", id="refresh-button", disabled=True),
+    )
+
+
+def create_manage_groups_nav_item(augur_manager):
+    """
+    Create the Manage Groups navigation item.
```
MoralCode: this documentation doesnt tell me anything new compared to the function name. comments should explain WHY we are doing something a certain way, or what something is for. The code already explains *what* it does

MoralCode (changes requested):
Skimmed through and left some feedback.

Given this PR is +600 - 300 in terms of diff (essentially taking the number of lines of code that get touched and doubling it), I'm a little unsure how this overall helps the codebase. It feels like we just replaced storing things in variables with wrapping them in parameter-less functions. Sure they are slightly more separated now, but i dont see how that helps actually refactor things to be more reusable or general-purpose

8Knot/pages/index/login_utils.py
EngCaioFonseca: I didn't modify any of the styling, just brought to the functions, but I'll change those as well.

8Knot/pages/index/login_utils.py
EngCaioFonseca: Some of the functions I've already made into more generic component, like the icon function, which works for three of the icons in the sidebar.
I'll work on making the functions more generic for other components as well.
The main idea of this refactoring was to make the code cleaner and remove the excessive nesting as it was before and therefore more readable and maintainable as well. I agree that the functions should be generic and reusable, that was the point, but the main one was to turn that whole nested chunk into a more "functional" one.
But it's fine with me working on making them more generic for each component as well.
Like looking into the all-in-one components.

EngCaioFonseca:
> Skimmed through and left some feedback.
> 
> Given this PR is +600 - 300 in terms of diff (essentially taking the number of lines of code that get touched and doubling it), I'm a little unsure how this overall helps the codebase. It feels like we just replaced storing things in variables with wrapping them in parameter-less functions. Sure they are slightly more separated now, but i dont see how that helps actually refactor things to be more reusable or general-purpose

I disagree with this pov, because the code overall is not only a function of lines of code added and subtracted.
The main thing being restructured/refactored in this PR is the index_layout, which went from having 400+ lines of code to 140. So it's a significant reduction in the structuring and the cleanliness of the code in the index_layout which is easier to read and modify, maintain, etc, and the additional lines of code are in utils file that contains the functions, which should also make it easier to find it and know where/what each functionality is/are.

cdolfi (commented):
overall looks good to me going to let @MoralCode take a look before merging

8Knot/pages/index/login_utils.py
```diff
@@ -0,0 +1,216 @@
...
+def create_nav_item(
+    item_id: str,
+    content: str,
+    disabled: bool = False,
+    href: Optional[str] = None,
+    external_link: bool = False,
+    target: Optional[str] = None,
+) -> dbc.NavItem:
+    """
+    Generic navigation item builder following Dash All-in-One Components pattern.
```
MoralCode, line 35: if this is meant to be an all in one component, it should probably be implemented as a class and be placed in its own file alongside our existing `VisualizationAIO` component.

8Knot/pages/index/login_utils.py
MoralCode: Alerts seem like another thing that are likely to be useful to implement in a reusable fashion. It seems lik this function hard-codes things like the alert title and body text that should be passed in as function args (or even better, refactored as an All In One Component Class)

8Knot/pages/index/login_utils.py
```diff
@@ -0,0 +1,216 @@
...
+
+    Returns:
+        html.Div containing the login disabled banner, or None if login is enabled
+    """
+    if not is_login_enabled():
+        return html.Div(
+            create_login_disabled_alert(),
+            className="login-banner-container",
+        )
+    return None
```
MoralCode, line 112: this seems like a really small function that encapsulates a very specific piece of functionality that - as far as i can tell - we likely only need in one place. Do we need a function for this?

8Knot/pages/index/login_utils.py
```diff
@@ -0,0 +1,216 @@
...
+        dbc.Popover component for login failure notifications
+    """
+    return dbc.Popover(
+        children="Login Failed",
+        body=True,
+        id="login-popover",
+        is_open=False,
+        placement="bottom-end",
+        target="nav-dropdown",
+    )
```
MoralCode, line 146: How is a popover different from the Alert we are using a few functions above this? Could this be another place where it may be helpful to reuse alerts?

8Knot/pages/index/login_utils.py
MoralCode: ok it looks like we do have a generic create alert function
https://github.com/oss-aspen/8Knot/pull/1007/files#diff-72ed40f2e8458b3ed27e11b3c9c0f731f8e0aa5d07bc58d4820ad77d255a4fc1R250

This specific one should probably be removed in favor of passing the parameters into that one

8Knot/pages/index/search_utils.py
```diff
@@ -229,3 +245,397 @@ def clean_repo_name(repo_name: str) -> tuple[str, str]:
...
+    ]
+
+
+def create_multiselect_styles():
+    """
+    Create the style configuration for the MultiSelect component.
+    Uses CSS variables from color.css and main_layout.css.
+
+    Note: DMC MultiSelect requires inline styles via the 'styles' prop.
+    This is a component-specific API, not regular inline styling.
```
MoralCode, line 399: thanks for documenting the reasons why these are inline and not in CSS!

8Knot/pages/index/search_utils.py
```diff
@@ -229,3 +245,397 @@ def clean_repo_name(repo_name: str) -> tuple[str, str]:
...
+    Returns:
+        dbc.Switch component for bot filtering
+    """
+    return dbc.Switch(
+        id="bot-switch",
+        label="GitHub Bot Filter",
+        value=True,
+        input_class_name="botlist-filter-switch",
+        className="bot-filter-switch",
+    )
```
MoralCode, line 536: If this is a function thats only used in `create_search_controls()`, it may be worth just merging it into that function with a comment

8Knot/pages/index/search_utils.py
```diff
@@ -229,3 +245,397 @@ def clean_repo_name(repo_name: str) -> tuple[str, str]:
...
+        disabled=disabled,
+        style=custom_style or {},
+        className=custom_class,
+    )
+
+
+# ===== LAYOUT COMPONENT FUNCTIONS =====
+
+
+def create_search_storage_components():
```
MoralCode, line 378: if this function is only used by `create_search_bar`, maybe it should be merged in with that function - unless we plan to reuse it.

8Knot/pages/index/search_utils.py
```diff
@@ -229,3 +245,397 @@ def clean_repo_name(repo_name: str) -> tuple[str, str]:
...
+                custom_class="icon-button",
+            ),
+            create_bot_filter_switch(),
+        ],
+        direction="horizontal",
+        className="search-controls-stack",
+    )
+
+
+def create_search_bar(initial_option):
```
MoralCode, line 567: Should we make our entire search bar into an All in One component? it seems like theres a lot of complexity there with all the data storage, bot filter switches, etc that only matter for the search bar. If we can encapsulate all this into one AIO component class, that will likely simplify this code a lot and allow people to just treat "the search bar" as one encapsulated thing (sorta blackbox-y) and not have to worry about whats inside it unless their change necessitates it

MoralCode (changes requested):
I see there has been some improvement with things like making a generic alert function with parameters so it can be reused.

If I were writing this PR, I'd still want to move more stuff into actual component classes. In addition to the benefits of reusability, AIO classes can also help move code to other parts of the codebase, decluttering these files further and making them more readable.
I also think proper AIO classes for major, but one-off components such as the searchbar, could also be beneficial for the above reason (encapsulation), and also because the searchbar is so complex, that giving it its own class (and own file) will allow us to keep all the various bits (data stores, toggle switches, etc) together, while also having more space to more thoroughly document it - especially since its a core component of the codebase.

cdolfi:
@EngCaioFonseca luk when this is ready for another round of reviews

EngCaioFonseca:
> @EngCaioFonseca luk when this is ready for another round of reviews

In line with what we discussed, would you be able to merge it and I'll keep working on the improvements from the issues we open outside of this PR?

cdolfi:
@EngCaioFonseca Squash commits and rebase and good to merge

EngCaioFonseca:
@cdolfi good to go.

cdolfi (approved):
lgtm

PR #1017, opened 2025-12-01
Search adaptive debounce

### LINKED ISSUE #1011: Advanced debounce timing - Search Bar
The current implementation of the debounce time in the app is a static debounce time regardless of the query length or the user behavior (i.e. keystroke speed)

A better implementation is to have an adaptive debounce time implementation, to account for the query length, so smaller queries (>3 chars) should fall under a specific time interval, and intermediate or larger queries should also have their own debounce time. Also, account for user keystroke speed, since if they're still typing fast, they are probably not done yet to fetch data.

EngCaioFonseca:
This PR implements an adaptive debouncing algorithm - strategy/logic to resolve issue #1011 

Brief summary description from Claude-Code, I find helpful to know apriori what all the changes are:

  Adds intelligent, query-length-based debouncing to the repository/organization search bar, significantly improving user experience and reducing unnecessary server requests.

  Key Features

  - Adaptive debounce timing: Short queries (1-3 chars) = 800-400ms, medium (4-5 chars) = 250ms, long (6+ chars) = 150ms
  - Smart immediate processing: Detects active typing and processes queries immediately when user is building up the search
  - Thread-safe state management: Handles concurrent Dash callbacks safely
  - Memory-conscious: Automatic cleanup of old queries after 5 seconds

  Technical Improvements

  - Performance: Eliminated redundant server fetches, saving 100-500ms per search
  - Code quality: Refactored 270-line monolithic function into 160-line orchestrator + 4 focused, reusable helpers
  - Error handling: Granular exception handling with graceful degradation
  - Testability: All search/formatting/selection logic now unit-testable

  Architecture

  AdaptiveDebounceManager (manages timing)
  ├── _perform_search() - Search orchestration
  ├── _format_search_results() - Result formatting
  ├── _handle_selected_options() - Selection management
  └── _get_fallback_selections() - Error recovery

  Impact

  - ✅ Better UX: No more premature searches while typing
  - ✅ Reduced server load: Intelligent request throttling
  - ✅ Maintainable: SOLID principles, 100% documented
  - ✅ Comprehensive error handling

### Generative AI disclosure
- This contribution was assisted or created by Generative AI tools.
    - What tools were used? Claude-Sonnet-4.5, Claude-Code, Coderabbit. 
    - How were these tools used?  Initial draft of the code / Prototyping / code review. 
    - Did you review these outputs before submitting this PR? Yes, many iterations of reviewing, improving and optimization of the code to make it cleaner, and aligned with best practices of SWE. One example was the optimization of the function for the adaptive debounce algorithm from a O(n) -> O(log n) -> O(1) which makes a substantial difference. 

## Summary by CodeRabbit

* **New Features**
  * Adaptive debouncing for search that adjusts delays by query length for more responsive results.
  * Visual search progress indicator shown by the search field to reflect ongoing searches.
  * Improved result formatting and fallback handling to ensure selected items appear and searches degrade gracefully.

* **Style**
  * New animations for the search progress indicator for smoother visual feedback.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/8Knot/pages/index/index_callbacks.py
+++ b/8Knot/pages/index/index_callbacks.py
@@ -89,6 +89,13 @@ class AdaptiveDebounceManager:
         Determine if a query should be processed based on adaptive debounce logic.
 
         Thread-safe: Uses locking to prevent race conditions in multi-threaded Dash environments.
+        Optimized with fast-path checks to minimize lock contention and reduce typing pauses.
+
+        Performance optimizations:
+        1. Fast-path lock-free check for duplicate queries (most common case)
+        2. Pre-compute expensive operations outside lock
+        3. Minimize time spent holding lock
+        4. Defer logging until after lock release
 
         Args:
             query: The current search query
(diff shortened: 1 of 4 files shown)


MoralCode (commented):
Why are we duplicating the functionality of existing dash debounce values?

I can see from the code comments that we want to make the debounce adaptive, which dash doesnt allow, but I want to dive deeper and ask: why do we need such advanced debounce in the first place?

The debounce we have set in dash (at least according to the code comments) seems to be 100ms. are our queries to the augur DB to check for repo and org names taking longer than that?

If performance for the search bar is the core problem, I'd want to know more about what testing was done to demonstrate that the application code is what needs to change here (as opposed to removing, tuning, or pre-computing one of the many caching or data storage layers that make up this application).

IMO if we have problems with these queries being so slow that dash callbacks are overlapping, I think we should figure out why they are slow and fix that.

One possible such fix that could be tried that I think would be better than trying to duplicate dash features and add thread locking to our callbacks, would be to pre-loading a list of all repo/org names/IDs into RAM (global variable, shouldnt need more than a dozen KB of RAM) when 8knot starts up and use that to near-instantlysuggest repos/orgs to the user as they type. Or even just load this data into a redis/valkey cache if theres issues with callbacks accessing a global variable in python.

8Knot/pages/index/index_callbacks.py
```diff
@@ -45,6 +48,396 @@
 # check if login has been enabled in config
 login_enabled = os.getenv("AUGUR_LOGIN_ENABLED", "False") == "True"
 
+
+# Adaptive debounce state manager
```
cdolfi, line 52: Is there a reason this class is in the index_callbacks vs the search _utils?

cdolfi:
@EngCaioFonseca I know the past we cached the repo/org information at start up, are we still doing that?

PR #1052, opened 2025-12-19
Search Async standalone solution. Solving UI freezes and increasing Client-side cache.

EngCaioFonseca:
This PR introduces the search async standalone solution. 
It increases the limits for the client-side cache, while also limiting the amount of repos/orgs are queried when user is typing and using the search bar. Before this solution, after every debounce was hit, and user queried the search bar, all 40k repos/orgs were queried at each keystroke which introduces large bottlenecks and UI freezes.

### Generative AI disclosure

    - What tools were used? Claude-Code
    - How were these tools used? Prototyping and ideation.
    - Did you review these outputs before submitting this PR? Yes. Manual review, modifications and thorough testing.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -6,6 +6,7 @@ http {
         server {
               listen 8080;
 	      access_log  /dev/null; # disables logging on every request
+              client_max_body_size 10M;         # allow larger request bodies for search results
               location / {
                 proxy_pass http://app-server:8080;
                 proxy_read_timeout 600s;        # 10 minutes - allow long-running Dash callbacks
--- a/8Knot/pages/index/index_callbacks.py
+++ b/8Knot/pages/index/index_callbacks.py
@@ -254,8 +254,8 @@ def dynamic_multiselect_options(user_in: str, selections, cached_options):
 
         # First, the search goes through the client-side cache if available
         if cached_options:
-            cache_matches = fuzzy_search(search_query, cached_options, threshold=search_threshold)
-            logging.info(f"Cache search found {len(cache_matches)} matches (threshold={search_threshold})")
+            cache_matches = fuzzy_search(search_query, cached_options, threshold=search_threshold, limit=1000)
+            logging.info(f"Cache search found {len(cache_matches)} matches (threshold={search_threshold}, limit=1000)")
 
         # Always also search server for comprehensive results (especially for longer queries)
         if len(search_query) >= 3:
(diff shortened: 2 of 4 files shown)


cdolfi (changes requested):
Getting errors when I try to use the search bar 

`<html> <head><title>413 Request Entity Too Large</title></head> <body> <center><h1>413 Request Entity Too Large</h1></center> <hr><center>nginx/1.29.1</center> </body> </html> <!-- a padding to disable MSIE and Chrome friendly error page --> <!-- a padding to disable MSIE and Chrome friendly error page --> <!-- a padding to disable MSIE and Chrome friendly error page --> <!-- a padding to disable MSIE and Chrome friendly error page --> <!-- a padding to disable MSIE and Chrome friendly error page --> <!-- a padding to disable MSIE and Chrome friendly error page -->`

Image

EngCaioFonseca:
> Getting errors when I try to use the search bar
> 
> `<html> <head><title>413 Request Entity Too Large</title></head> <body> <center><h1>413 Request Entity Too Large</h1></center> <hr><center>nginx/1.29.1</center> </body> </html> <!-- a padding to disable MSIE and Chrome friendly error page --> <!-- a padding to disable MSIE and Chrome friendly error page --> <!-- a padding to disable MSIE and Chrome friendly error page --> <!-- a padding to disable MSIE and Chrome friendly error page --> <!-- a padding to disable MSIE and Chrome friendly error page --> <!-- a padding to disable MSIE and Chrome friendly error page -->`
> 
> Image

Issue is resolved now. Could you please test it again?

cdolfi (commented):
lgtm, will let @MoralCode give it a review once hes back in office

nginx.conf
```diff
@@ -6,6 +6,7 @@ http {
         server {
               listen 8080;
 	      access_log  /dev/null; # disables logging on every request
+              client_max_body_size 10M;         # allow larger request bodies for search results
```
MoralCode, line 9: How concerned are we about the potential security impacts of this?

Example:

I, an evil attacker, know that this web server is publicly accessible and can send responses up to 10MB in response to a search request.

I have someone's IP whom i would like to DDoS.

I send queries to 8knot that mimic search queries that return a lot of data, but i spoof my IP address to the IP of my victim. Therefore i send a tiny packet (easily under a few KB) with my search query and get 10M of data back, but the spoofed IP causes the response to go to my victim, flooding them with orders of magnitude more data than my initial request, effectively amplifying the power of my attack.

This is called a DDoS amplification attack and is described here: https://www.microsoft.com/en-us/security/blog/2022/05/23/anatomy-of-ddos-amplification-attacks/

Should we be worried about this?

MoralCode: actually... probably not since TCP mitigates this kind of source IP spoofing as explained later in the article... carry on.

MoralCode:
overall i think i generally agree with these changes though! can probably give it a test too if that would help

cdolfi:
https://github.com/oss-aspen/8Knot/blob/dev/8Knot/pages/affiliation/visualizations/org_associated_activity.py this is a good example of how this should be formatted (just one less layer of dbc.col)

MoralCode:
@cdolfi i think you have the wrong PR. This one doesnt contain dbc.col

EngCaioFonseca:
Just went through everything and tested it again. It seems to be ready to go if you're guys are good with it. Lmk if there's any other modification necessary.

cdolfi:
@MoralCode yeah you're right that comment was to a different pr, from my side this lgtm

cdolfi (approved):
lgtm

MoralCode (approved):
LGTM, solid performance gain.


the difference is most noticeable when typing a long name in the searchbar (ansible, kubernetes). before: when you type normally on first load it takes a while (but not unacceptable) to load, and then as you delete characters it almost feels like it gets slower the shorter it gets.

With the new one its pretty snappy all around.

Id wonder if, with the new limits, we even need the 10M limit

PR #1060, opened 2026-01-10
Use math.log10 instead of math.log for logarithmic plots

### LINKED ISSUE #986: Use `math.log10` instead of `math.log` for logarithmic plots
Project velocity can use logarithmic axis, which is quite useful for spreading out projects, but it uses `math.log` https://github.com/oss-aspen/8Knot/blob/abf4cc335f267ace6d5c37a0e149d9e742d9cb25/8Knot/pages/chaoss/visualizations/project_velocity.py#L358-L361 https://github.com/oss-aspen/8Knot/blob/abf4cc335f267ace6d5c37a0e149d9e742d9cb25/8Knot/pages/chaoss/visualizations/project_velocity.py#L376 However I think that for humans it's easier to reason in terms of decades than it is of...powers of the Euler constant.  If I a repo has a number of log10(commit) between 3 and 4 I can immediately tell "oh, that repo has ~thousands of commits", if it's between 1 and 2 I can easily guess "that one instead has only a few dozens", etc...while I have much more difficulties making the same quick guesses with the natural logarithm.  A variation is to plot in log10 scale but use the linear values on the axes ticks: 1, 10, 100, 1000, etc..., which makes things even easier to read.

Sukuna0007Abhi:
Fixes #986

### What I changed
Changed math.log (natural log, base e) to math.log10 (base 10) in the project velocity visualization. Also made the axes show actual values like 1, 10, 100, 1000 instead of log values.

### Why
As @giordano mentioned in the issue, base-10 logarithms are way easier to understand. When you see a value between 3 and 4, you immediately know "that's thousands of commits" rather than having to calculate powers of e in your head.

### What changed

- All math.log calls now use math.log10
- The graph axes show real values (10, 100, 1000, etc.) using Plotly's log_x and log_y
- Hover tooltips show both the raw value and log10 value so you can see the comparison
- This implements both the main fix and the "variation" suggested in the issue for showing linear tick values.

@cdolfi Thanks for tagging me on this one!

Pls review @cdolfi @MoralCode

### DIFF
--- a/8Knot/pages/chaoss/visualizations/project_velocity.py
+++ b/8Knot/pages/chaoss/visualizations/project_velocity.py
@@ -299,10 +299,10 @@ def process_data(
     # replace all nan to 0
     df_consolidated.fillna(value=0, inplace=True)
 
-    # log of commits and contribs if values are not 0
-    df_consolidated["log_num_commits"] = df_consolidated["Commit"].apply(lambda x: math.log(x) if x != 0 else 0)
+    # log10 of commits and contribs if values are not 0 (base-10 for intuitive decade-based reasoning)
+    df_consolidated["log_num_commits"] = df_consolidated["Commit"].apply(lambda x: math.log10(x) if x != 0 else 0)
     df_consolidated["log_num_contrib"] = df_consolidated["num_unique_contributors"].apply(
-        lambda x: math.log(x) if x != 0 else 0
+        lambda x: math.log10(x) if x != 0 else 0
     )
 
     # column to hold the weighted values of pr and issues actions summed together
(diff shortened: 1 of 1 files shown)


cdolfi (changes requested):
please only change from log to log 10 in this pr to stay within the scope of the issue

Sukuna0007Abhi:
> please only change from log to log 10 in this pr to stay within the scope of the issue

Hi ma'am @cdolfi actually in the issue #986, @giordano suggested "variation"  for showing linear tick values.

So, that's why I did both math.log to math.log10 and this variation...

The graph axes show real values (10, 100, 1000, etc.) using Plotly's log_x and log_y &
Hover tooltips show both the raw value and log10 value so you can see the comparison...

If you think this is not needed then I will revert it pls review one time more..

PR #1061, opened 2026-01-10
Fix organization core contributors styling issue 1005

### LINKED ISSUE #1005: Fix organization core contributors styling
As many toggles that can fit on the same line should 

Image

Sukuna0007Abhi:
- Consolidated controls from 3 rows to 2 rows
- Moved email filter toggles to first row with input controls
- Added responsive breakpoints (xs=12, sm=auto) for mobile support
- First row: 2 inputs + 2 toggles (stacks on mobile, one row on desktop)
- Second row: date picker
- Added proper spacing with Bootstrap classes

This puts as many toggles on the same line as requested while maintaining good UX on all screen sizes.

Fixes #1005


cdolfi (changes requested):
8knot is not intended to support mobile, please keep to the width and classnName conventions consistent to the other visualizations

PR #1064, opened 2026-01-10
Added .gitattributes to resolve #1062

### LINKED ISSUE #1062: Add .gitattributes for defining how to handle line endings

AvneeshYadav02:
Added a .gitattributes file to handle line endings for python files.
verified with pre-commit.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.py text eol=lf


MoralCode:
Hello, thanks for the PR.

Would you be able to join the [CHAOSS Slack](https://chaoss.community/kb-getting-started/)  #wg-augur-8knot channel? i had some questions for you and the contributors of the other PRs for this same thing

PR #1063, opened 2026-01-10
Add .gitattributes for consistent line ending handling

AndyBodnar:
## Summary
- Adds `.gitattributes` file to ensure Python files always use LF line endings
- Prevents CRLF issues when Windows contributors submit changes
- Addresses the root cause mentioned in #1061

## Changes
- Created `.gitattributes` with rules for:
  - Python files (`*.py`) - always LF
  - Shell scripts (`*.sh`) - always LF  
  - Docker/config files - always LF
  - Batch files (`*.bat`, `*.cmd`) - always CRLF (Windows-native)
  - Binary files - no modification

## Test plan
- [x] Verify file is correctly formatted
- [ ] Existing contributors can verify their local line endings normalize correctly

Fixes #1062

PR #1065, opened 2026-01-11
Add .gitattributes to enforce LF line endings

### LINKED ISSUE #1062: Add .gitattributes for defining how to handle line endings

Sukuna0007Abhi:
- Set default text handling to auto-normalize line endings
- Explicitly enforce LF line endings for Python files
- Add LF enforcement for other text files (md, yml, sql, css, etc.)
- Mark binary files appropriately
 Fixes #1062

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,18 @@
+# Set default behavior to automatically normalize line endings
+* text=auto
+
+# Python files should always use LF line endings
+*.py text eol=lf
+
+# Other text files in this repository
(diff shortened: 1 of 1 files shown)


MoralCode:
@Sukuna0007Abhi @nexpectArpit do you two know each other? cant say ive ever seen a PR on a fork by the submitter force pushed to by a totally different account before

Sukuna0007Abhi:
Closing it as #1063 will be done for it

PR #1086, opened 2026-02-14
Enable heatmap queries with materialized views

### LINKED ISSUE #803: Materialized view for cntrib_by_file query

EngCaioFonseca:
- Fix cache_facade.py: Add missing cache_conn.commit() call
- Re-enable 3 query tables in db_init.py
- Re-enable imports in index_callbacks.py
- Update queries to use materialized views for better performance:
  - cntrb_per_file_query: Use augur_data.explorer_cntrb_per_file
  - pr_files_query: Use augur_data.explorer_pr_files
  - repo_files_query: Use augur_data.explorer_repo_files

### Generative AI disclosure

    - How were these tools used?  Draft, prototype.


8Knot/cache_manager/db_init.py
```diff
@@ -256,43 +256,43 @@ def _create_application_tables() -> None:
...
+            )
+            """
+        )
+        logging.warning("CREATED cntrb_per_file_query TABLE")
+
+        cur.execute(
+            """
+            CREATE UNLOGGED TABLE IF NOT EXISTS pr_file_query(
+                file_path text,
+                pull_request int,
```
MoralCode, line 276: this variable was renamed from pull_request_id. is there a reason for that? it seems like it is less clear what it is for now than before

8Knot/pages/index/index_callbacks.py
```diff
@@ -23,9 +23,10 @@
 from queries.user_groups_query import user_groups_query as ugq
 from queries.pr_response_query import pr_response_query as prr
 
-# from queries.cntrb_per_file_query import cntrb_per_file_query as cpfq - codebase page disabled
-# from queries.repo_files_query import repo_files_query as rfq - codebase page disabled
-# from queries.pr_files_query import pr_file_query as prfq - codebase page disabled
+# Codebase page queries - enabled for heatmap visualizations
```
MoralCode: There are non-contribution query imports below this. This comment may be misleading

8Knot/queries/cntrb_per_file_query.py
```diff
@@ -30,19 +30,14 @@ def cntrb_per_file_query(self, repos):
...
-                    pull_requests pr,
-                    pull_request_files prf,
-                    pull_request_reviews prr
+                    augur_data.explorer_cntrb_per_file
                 WHERE
-                    pr.pull_request_id = prf.pull_request_id AND
-                    pr.pull_request_id = prr.pull_request_id AND
-                    pr.repo_id in %s
-                GROUP BY prf.pr_file_path, pr.repo_id
+                    repo_id in %s
```
MoralCode, line 40: this looks like we are potentially refactoring a lot of this query to rely on the materialized view. Is that what is happening here/is there an explanation for why this is needed in this PR rather than being in phase 2?

MoralCode (commented):
I still need to test this but had some questions about certain aspects of this change, mostly looking for the underlying reasons why we are changing the queries when presumably they worked before the heatmap was disabled

8Knot/cache_manager/db_init.py
cdolfi: It doesnt look to me that there is a change here. The # just needed to be taken out to re-enable

8Knot/pages/index/index_callbacks.py
EngCaioFonseca: I'll remove the comment.

8Knot/queries/cntrb_per_file_query.py
EngCaioFonseca: Yes, that's pretty much it. Just refactoring now to use the materialized views instead. 
This is being implemented here as the original issue was to re-enable the heatmaps after materialized views were created for them, like the title suggests. So, this PR contains both re-enabling the heatmaps and the refactoring of the query to use the mat. views.

cdolfi: yes the first phase of this is to change the queries to materialized views. They were disabled originally bc they were taking so long and crashing the replica

cdolfi (changes requested):
https://github.com/oss-aspen/8Knot/tree/dev/docs/materialized_views - the materialized view contents should be added to the docs and a note similar to this one should get added to each of the queries. 

Also the codebase page needs to get added to the navbar so the heatmaps are back on the application and we can see if the graphs work with the materialized views before the rest of the refactor

EngCaioFonseca:
Sounds good. I'll make the adjustments addressing the reviews.

8Knot/pages/index/index_layout.py
```diff
@@ -74,6 +74,12 @@
                     active="exact",
                     className="nav-link nav-link--visualization",
                 ),
+                dbc.NavLink(
```
cdolfi: This should be added to the side nav with all of the other visualization pages

cdolfi (changes requested):
Im getting errors on start up and cant fully test if this works. From what I can see its the repo dropdown that is causing the issues. It should work like the one in repo overview page https://github.com/oss-aspen/8Knot/blob/dev/8Knot/pages/repo_overview/repo_overview.py
Screenshot 2026-02-18 at 5 58 19 PM

8Knot/pages/index/index_layout.py
EngCaioFonseca: Okay, makes sense. I actually forgot to ask today on the meeting, because I placed it on the top navbar, but wasn't sure about it, and I was gonna ask if it was supposed to be in the side nav. Well, now we know.

8Knot/cache_manager/db_init.py
MoralCode: github is also flagging that `pull_request_id int,` got changed to `pull_request int,`

MoralCode:
> Also the codebase page needs to get added to the navbar so the heatmaps are back on the application and we can see if the graphs work with the materialized views before the rest of the refactor

So the wording here may have been misleading, but since this is a visualization, the link should go in the sidebar of the visualizations page, not in the top navbar.

Since the disclosure indicates AI was used here, Do we know whether this was a human miscommunication, or an AI hallucination?

8Knot/pages/index/index_layout.py
cdolfi: Top nav bar is just for differentiating between welcome page and visualizations as a whole. The side nav bar is for the visualization pages and visualizations

EngCaioFonseca:
> > Also the codebase page needs to get added to the navbar so the heatmaps are back on the application and we can see if the graphs work with the materialized views before the rest of the refactor
> 
> So the wording here may have been misleading, but since this is a visualization, the link should go in the sidebar of the visualizations page, not in the top navbar.
> 
> Since the disclosure indicates AI was used here, Do we know whether this was a human miscommunication, or an AI hallucination?

Not AI, the choice was mine, so just good ol' human miscommunication, but it makes sense, for sure it didn't feel totally right in there. Well, now it's clear anyways. But yeah, It wasn't totally clear to me due to the wording as well.

EngCaioFonseca:
> Im getting errors on start up and cant fully test if this works. From what I can see its the repo dropdown that is causing the issues. It should work like the one in repo overview page https://github.com/oss-aspen/8Knot/blob/dev/8Knot/pages/repo_overview/repo_overview.py Screenshot 2026-02-18 at 5 58 19 PM

Okay, it should be fixed now.  Mantine core requires the options to be strings instead of ints, so, I've readjusted for that. Could you have another go at it?

8Knot/cache_manager/db_init.py
```diff
@@ -256,43 +256,43 @@ def _create_application_tables() -> None:
...
-        #     )
-        #     """
-        # )
-        # logging.warning("CREATED cntrb_per_file_query TABLE")
-
-        # cur.execute(
-        #     """
-        #     CREATE UNLOGGED TABLE IF NOT EXISTS pr_file_query(
-        #         file_path text,
-        #         pull_request_id int,
```
MoralCode, line 276: This can be changed in the materialized view query instead of here

8Knot/cache_manager/db_init.py
```diff
@@ -256,43 +256,43 @@ def _create_application_tables() -> None:
...
-        #         repo_name text,
-        #         repo_path text,
-        #         rl_analysis_date text,
-        #         file_path text,
-        #         file_name text
-        #     )
-        #     """
-        # )
-        # logging.warning("CREATED repo_files_query TABLE")
+        # Codebase page tables - enabled for heatmap visualizations
```
MoralCode, line 259: this comment isnt needed. there was only a comment here to say why this block was commented

8Knot/cache_manager/db_init.py
MoralCode: This can be changed in the materialized view query to match what is here

8Knot/pages/codebase/visualizations/cntrb_file_heatmap.py
```diff
@@ -145,9 +145,36 @@ def repo_dropdown(repo_ids):
...
+    # Find first repo with valid metadata (non-empty repo_name and repo_path)
+    default_repo = None
+    for repo_id in repo_ids:
+        try:
+            df = cf.retrieve_from_cache(tablename=rfq.__name__, repolist=[repo_id])
+            if not df.empty:
+                repo_name = df["repo_name"].iloc[0] if "repo_name" in df.columns else None
+                repo_path = df["repo_path"].iloc[0] if "repo_path" in df.columns else None
+                # Skip repos with NULL or empty metadata
+                if repo_name and repo_path and repo_name != "None" and repo_path != "None":
```
MoralCode: this is a lot of logic to have in the callback.

Can we not have the query include like a NULLS LAST or something to not get a null in what we seem to be returning as the like "default value"?

8Knot/pages/codebase/visualizations/cntrb_file_heatmap.py
```diff
@@ -160,15 +187,22 @@ def repo_dropdown(repo_ids):
     background=True,
 )
 def directory_dropdown(repo_id):
+    # Handle None repo_id case
+    if repo_id is None:
+        logging.warning(f"{VIZ_ID} DROPDOWN - repo_id is None")
+        return ["Top Level Directory"], "Top Level Directory"
```
MoralCode: what is the value "Top Level Directory" representing/meaning here?

docs/materialized_views/explorer_repo_files.sql
```diff
@@ -0,0 +1,21 @@
+/* This is the SQL query that populates the explorer_repo_files materialized view*/
+SELECT
+    rl.repo_id AS id,
```
MoralCode, line 3: This does not match the query you provided me to install on our prod machine (which happened to already exist in augur i'm pretty sure).

that query just has this column be called repo_id

8Knot/pages/index/index_components.py
```diff
@@ -204,6 +204,25 @@ def create_sidebar_navigation():
...
+                    NavItemAIO(
+                        "Contributor File Heatmap",
+                        "/codebase#cntrb-file-heatmap",
+                    ),
+                    NavItemAIO(
+                        "Reviewer File Heatmap",
+                        "/codebase#reviewer-file-heatmap",
+                    ),
+                ],
+                "/assets/repo_overview.svg",
```
MoralCode: I just pushed a commit to change this to the codebase svg file i got from figma

MoralCode:
Overall I think the reason we are struggling to review PRs like this is that its hard for us to know whats in your head and each of us has a different set of internal questions that come up as we read the code, and if those arent answered through intuition (i.e. thinking about how we would have solved the problem, or why things were done that way) then it becomes review feedback that can drag on for a while since we all need to be aligned in order for something to merge.

Example: Its hard for me to understand what you mean by "it fixes the nulls issue" unless there's more details (what stack trace or error/symptoms were you seeing, what commit was that - can i go back and reproduce it and hack on it myself if i think there's a better way, or better yet, have your explanation outline all the simpler fixes you tried first and why they didn't work)

Other than the small suggestions of moving changes to the mat views, etc I think my main understanding mismatch with you here is surrounding the null issue. Your solution may be right, but i don't have enough context to add to my own mental model of how things work so the only thing that is currently available in my brain for feedback is "thats a lot of code to do filtering in a callback when really we should probably be handling that at a query level".

I think the thing that matters most here is that: its not just having a solution to all the problems that makes a good PR, but also having the information alongside it to help people understand why it was made. )

The disclosure that AI was used to draft and prototype the solution sorta adds to that information, but in a backwards way. As someone who is somewhat skeptical of the abilities of GenAI to write good code with the same attention to design as I would, I'm somewhat less confident that there aren't extraneous changes that have been added besides what you intended (you also weren't specific about  how you reviewed the AI output and/or what you changed or removed).

I think this has come a long way - i still need to retest it (probably after we figure out where the mismatch between my brain and yours is when it comes to this null issue), but it looks like this is coming along okay.

8Knot/pages/codebase/visualizations/cntrb_file_heatmap.py
```diff
@@ -148,7 +148,15 @@ def repo_dropdown(repo_ids):
         entry = {"value": str(repo_id), "label": augur.repo_id_to_git(repo_id)}
         data_array.append(entry)
 
-    return data_array, str(repo_ids[0])
+    # Select first repo with valid cached data as default
+    default = str(repo_ids[0])
+    for repo_id in repo_ids:
+        df = cf.retrieve_from_cache(tablename=rfq.__name__, repolist=[repo_id])
```
MoralCode: im skeptical of the performance of doing retrieve_from_cache (which i assume does db calls) in a for loop - is there not a way we can retrieve the data and then iterate over each result and check/remove problematic IDs or extract the ones that are known good? 

The latency to even hit the cache when run this many times is probably going to add up really fast

MoralCode (changes requested):
curious about the performance here

8Knot/pages/codebase/visualizations/cntrb_file_heatmap.py
EngCaioFonseca: you're right, it makes sense! going with the retrieve_from_cache inside the loop is not the best alternative here, performance wise. I brought the retrieve_from_cache outside the loop.

8Knot/pages/codebase/visualizations/contribution_file_heatmap.py
```diff
@@ -430,7 +441,7 @@ def pr_per_directory_value(directory, df_file):
     df_dynamic_directory = df_file[df_file["file_path"].str.startswith(directory)]
 
     # test if there is any pull requests in the directory
-    if df_dynamic_directory.pull_request_id.isnull().all():
+    if df_dynamic_directory.pull_request.isnull().all():
```
MoralCode, line 444: this change mirrors the change that was made to the queries themselves. can we revert both of them so it still matches but doesnt remove useful information from the name?

MoralCode (changes requested):
overall this is very close, just one small change to make. ill try and test this today to make sure it runs

good catch with renaming the fields that no longer exist in the query - i just think this should be solved by setting the queries back to how they were originally so there arent unrelated changes in the diff

8Knot/pages/codebase/visualizations/contribution_file_heatmap.py
EngCaioFonseca: Could this be done in another pr instead of this one? the alias pull_request_id as pull_request was already in origin/dev so I only adjust it to make sure it works on the viz. We can go in either direction: either we fix it on a smaller follow up pr, or fix everything back on this one. IMO since this one already took long enough, and achieves its goal, we should leave it for a minor follow up pr.

cdolfi (commented):
Ran it locally and it works, make sure to open an issue for the visualization refactor and updating it to the current app design

EngCaioFonseca:
Done. 

Issue: #1099

8Knot/pages/codebase/visualizations/contribution_file_heatmap.py
```diff
@@ -393,6 +456,10 @@ def df_file_clean(df_file: pd.DataFrame, df_file_pr: pd.DataFrame):
     df_file.drop(["repo_id"], axis=1, inplace=True)
     df_file_pr.drop(["repo_id"], axis=1, inplace=True)
 
+    # Rename pull_request column to pull_request_id for consistency
+    if "pull_request" in df_file_pr.columns:
+        df_file_pr = df_file_pr.rename(columns={"pull_request": "pull_request_id"})
+
```
MoralCode, line 462: We dont need to rename these columns again we can just undo where they were renamed the first time (it happened in this PR when the queries were uncommented)

MoralCode (changes requested):

8Knot/pages/codebase/visualizations/contribution_file_heatmap.py
MoralCode: could it? sure. Should it? IMO no.

pull request_id as pull_request was something that this PR has modified as well https://github.com/oss-aspen/8Knot/pull/1086/changes#diff-d75d3c646fe5ed269406728261f4faf9d450ce0e5a93ba7019d593400276836bL35

The intent of this PR is to be a minimal change to re-enable the heatmaps. I feel like changing the queries and then introducing changes to un-do those first changes is adding a lot more code than necessary for a simple scoped PR

8Knot/queries/cntrb_per_file_query.py
MoralCode: i have a feeling the true root cause of this is https://github.com/chaoss/augur/issues/3736, but thats a bigger issue than can be solved now

MoralCode (approved):
I give up trying to be perfectionist about not changing the naming of db tables and things in this PR. lets just ship this, it seems to work well enough on my machine.

MoralCode:
gonna merge this since cali has previously said shes okay with it merging once i was ok with it

PR #1100, opened 2026-03-04
Increasing soft time limit on worker-callback 

### LINKED ISSUE #1083: Graphs are crashing when waiting for data to load
More investigation and experimentation is needed to see how 8knot is behaving when loading larger repos: 

pytorch/pytorch is what is causing this. I would be curious to see the loading of repos of similar size 

Image

Even if the graphs load check to make sure its with the data of the repository selected vs chaoss

EngCaioFonseca:
This PR aims to resolve issue #1083 on the larger repos loading data issue.
Loading data for larger repos was taking significantly more time than the worker-callback softime limit, after that a reset/reload command was also triggered because of that, pointing back to the org Chaoss by default.
Increasing the time limits on the worker-callback aligned and in sync with the already increased worker-query time limits should resolve the issue. 

### Generative AI disclosure

Please select one option:

- [x] This contribution was NOT assisted or created by Generative AI tools.

    - What tools were used? 
    - How were these tools used? 

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -67,7 +67,7 @@ services:
     build:
       context: .
       dockerfile: ./docker/Dockerfile
-    command: ["celery", "-A", "app:celery_app", "worker", "--loglevel=INFO", "--concurrency=1", "--time-limit=300", "--soft-time-limit=240"]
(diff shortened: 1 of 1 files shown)


MoralCode (approved):
Simple enough code - lets release this asap

PR #1107, opened 2026-03-31
Self merge graph

EngCaioFonseca:
This PR adds the Self Merge Visualization into the app. 

ref: https://github.com/chaoss/wg-data-science/issues/182

### Generative AI disclosure

Please select one option:

    - What tools were used? Claude Code
    - How were these tools used? Initial Draft.
    - Did you review these outputs before submitting this PR? Yes.


8Knot/cache_manager/db_init.py
```diff
@@ -195,6 +195,22 @@ def _create_application_tables() -> None:
         )
         logging.warning("CREATED prs TABLE")
 
+        cur.execute(
+            """
+            CREATE UNLOGGED TABLE IF NOT EXISTS self_merge_rate_query(
```
cdolfi: I asked this question in the notebook as well, Why does this need to be a new query vs and edit to the pr query?

8Knot/pages/contributions/contributions.py
```diff
@@ -30,6 +31,10 @@
             dbc.Col(gc_pr_over_time, xl=10),
             className="visualization-row",
         ),
+        dbc.Row(
+            dbc.Col(gc_self_merge_rate, xl=10),
```
cdolfi: move to after the staleness graph

8Knot/queries/self_merge_rate_query.py
```diff
@@ -0,0 +1,58 @@
...
+    Pull requests that have merged, with merger contributor id from the latest
+    merge event
+    """
+    logging.warning(f"{self_merge_rate_query.__name__} COLLECTION - START")
+
+    if len(repos) == 0:
+        return None
+
+    query_string = """
+        SELECT
```
cdolfi: This is different than what is in the notebook

cdolfi (commented):
Before fully reviewing my main question is around the query, why does it need its own vs an edit to the  pr query? 

The query and graph code used in the notebook should be the same as the one in this PR. If there are needed edits it should be in a PR to the notebook first so it is consistent and can be properly reviewed

8Knot/cache_manager/db_init.py
EngCaioFonseca: My first goal/idea was to do exactly that, add/edit into prs_query, however it seems that having an independent new query is the best choice here due to these reasons:
* The SQL in the notebook has/adds LEFT JOIN pull_request_events pre ON pr.pull_request_id = pre.pull_request_id AND pre.action= 'merged'. If a PR has more than one merged row in Augur (which can happen), that join will duplicate rows. 
* If we choose to go and edit prs_query with this JOIN, it could break pr_over_time, pr_staleness, pr_first_response, and so on, other visualizations which reads from that cache table. So a dedicated query avoids this.

That's my line of thoughts on why a dedicated query is a better fit in this case.

cdolfi: `SELECT
    r.repo_id,
    r.repo_name,
    pr.pull_request_id,
    pr.pr_src_number,
    left(pr.pr_augur_contributor_id::text, 15)  AS cntrb_id,
    pr.pr_created_at                             AS created_at,
    pr.pr_closed_at                              AS closed_at,
    pr.pr_merged_at                              AS merged_at,
    -- Only populated for merged PRs (NULL for open/closed-unmerged)
    left(pre.cntrb_id::text, 15)                AS merger_cntrb_id
FROM repo r
JOIN pull_requests pr
    ON r.repo_id = pr.repo_id
LEFT JOIN pull_request_events pre
    ON  pr.pull_request_id = pre.pull_request_id
    AND pre.action = 'merged'
WHERE r.repo_id IN ({repo_statement})
  AND pr.pr_created_at < now()
  AND (pr.pr_closed_at  < now() OR pr.pr_closed_at  IS NULL)
  AND (pr.pr_merged_at  < now() OR pr.pr_merged_at  IS NULL)
ORDER BY pr.pr_created_at;`

Would this query work? I think its only adding a single additional field to the pr_query vs a whole new one. Might make sense to atleast add some indexes/turn it into a materialized view

EngCaioFonseca: It works, but still doesn't stop duplication overall. So, I'm testing this path, but adding something to be able to get only distinct rows, to avoid the duplicates. I'll commit soon if it works

EngCaioFonseca: Okay, done. I added the DISTINCT ON on the query to avoid duplications, which should work (final testing it now). It still gonna make it heavier on the prs_query side rather than having a dedicated one, but hopefully won't make that much of a difference.

cdolfi: It will make it more complicated but that is okay. Would rather that than duplicated collection of the same data. Please make these changes in the notebook first and I can better review there. Then on this side it will be an easy review

cdolfi: Actually... There should not be any duplication and Distinct can be pretty costly. Can you elaborate on what you are seeing to need to deduplicate?

EngCaioFonseca: Screenshot 2026-04-01 at 14 45 43

Screenshot 2026-04-01 at 14 45 58

These were the last modifications. 
I removed the DISTINCT ON from the query as it was taking way longer to get done, and moved the duplication handling into python. Which I'll also modify in the notebook. 
I've tested it here, and seems to work well.

cdolfi: What I am trying to say is there should only be one row per pull request so if there is duplicates something is wrong. Would you be able to elaborate on how you found duplicates?

cdolfi: cc @MoralCode

EngCaioFonseca: It was more of a security redudancy, precaution, in case there were multiple cases of the new variables introduced in the query. But I didn't find any duplicates anyways. If we're sure there's no possible duplicates, I can implement without any handling of it no problem. I'll do it now.

And yes, I agree, the distinct on does add a lot of weight into it, so that's why I moved it into a python handling of the case. But I'll just drop it, so all should be good in any case.

EngCaioFonseca: Done.

cdolfi: Yeah there should not be any duplicates. When it comes to query or graph changes that should be reviewed at the notebook step so tag me in that PR when its ready

EngCaioFonseca: @cdolfi https://github.com/chaoss/wg-data-science/pull/195 Query on the notebook side has now been updated and match the implementation in here for the prs_query, everything else already matches (data processing and graphs)

cdolfi:
FYI this will need to be tested, the graph is not building. I am assuming from reading the code that these three plots would be on the same graph or show one after another. Either way please implement similar to Drive-Thru Contributions Per Quarter where there are 2 graph views

cdolfi:
Also add in "Pull Request" to each of the titles

cdolfi (approved):
lgtm

PR #1108, opened 2026-03-31
Search UI revisited

### LINKED ISSUE #1097: Enter/Return key doesn't work to complete search, unintuitive placement of icon to click
**Describe the bug**

This could be considered two bugs:

- As a user, I should be able to search and hit Enter or Return to complete the query (have it go from grey to blue).
- As as user, I shouldn't have to click an icon on the left after completing a query on the right. This is unintuitive.

**To Reproduce**
1. Go to https://metrix.chaoss.io
1. Click on the hamburger icon 🍔 
1. Click the "x" to remove the default `org:chaoss` query.
1. Type a new query. I'm using "dataverse".
1. Scroll down until you find the right one. Hit Enter/Return to select it.
1. Note that the result is still grey.
1. Wonder what you did wrong. 🤔 Why can't I see the data for the repo? 😢 

**Expected behavior**

I expect the item (a repo in this case) to be selected.

**Screenshots**

## scrolling to find the right repo

Image

## result is still grey

Image

**Additional context**

See https://chaoss-workspace.slack.com/archives/C0226ELG6R4/p1772128187122029

Related:

- https://github.com/oss-aspen/8Knot/issues/739

Thank you to @MoralCode for pointing me toward the answer! ❤️ The Enter/Return key doesn't work. You have to click the search icon. This will turn the query blue (which I didn't know was important), like this:

Image

Once the query is blue, you can click "Repo Overview", for example, and see the data for the repo you want, like this:

Image

EngCaioFonseca:
This PR adds a few features to improve the Search bar UI/UX based on Issue #1097 

### Generative AI disclosure

Please select one option:

    - What tools were used? Claude Code, Coderabbit
    - How were these tools used? Initial draft and Prototype
    - Did you review these outputs before submitting this PR? Yes

### DIFF
--- a/8Knot/assets/search_keyboard.js
+++ b/8Knot/assets/search_keyboard.js
@@ -1,0 +1,42 @@
+/**
+ * Cascading Enter key handler for the search bar.
+ *
+ * When the Mantine MultiSelect dropdown is OPEN, Enter selects the
+ * highlighted item (Mantine's default behavior — we do nothing).
+ *
+ * When the dropdown is CLOSED, Enter triggers the search by clicking
+ * the search button, which fires the existing Dash callback chain.
+ */
+document.addEventListener("DOMContentLoaded", function () {
+    // Use MutationObserver to attach the handler once the MultiSelect input exists
+    var observer = new MutationObserver(function () {
+        var input = document.querySelector("#projects input");
+        if (!input) return;
+
+        // Stop observing once we've found the input
+        observer.disconnect();
+
+        input.addEventListener("keydown", function (e) {
+            if (e.key !== "Enter") return;
+
+            // Check if the Mantine dropdown is currently open
+            var dropdown = document.querySelector(
+                ".mantine-MultiSelect-dropdown"
+            );
+            if (dropdown) {
+                // Dropdown is open — let Mantine handle the Enter key
+                // to select the highlighted item
+                return;
+            }
+
+            // Dropdown is closed — trigger search
+            e.preventDefault();
+            var searchButton = document.getElementById("search-button");
+            if (searchButton) {
+                searchButton.click();
+            }
+        });
+    });
+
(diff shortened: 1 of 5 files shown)


cdolfi (commented):
Reading through this seems like it should be split into atleast 2 PR: One for color styling changes, one for button change, and one for the Cascading Enter key handler for the search bar.

8Knot/assets/search_keyboard.js
```diff
@@ -0,0 +1,44 @@
...
+                ".mantine-MultiSelect-option"
+            );
+            if (options.length > 0) {
+                // Dropdown has options — let Mantine handle Enter
+                // to select the highlighted item
+                return;
+            }
+
+            // Dropdown is closed — trigger search
+            e.preventDefault();
```
MoralCode, line 35: we shouldnt be trying to look into the children to decide what to do in a parent event handler - this is backwards from the natural cascading effect of event handlers.

instead we should allow the menu items to handle their own key events (we already do this somewhere i think since pressing enter on one option already works), and simply e.PreventDefault() after doing that so it doesnt bubble up to the search bar. Then, once the menu is closed, the search bar should be in focus and a subsequent enter will hit this event handler, eliminating the need for this logic and simplifying it down to just calling the search function

8Knot/assets/search_keyboard.js
```diff
@@ -0,0 +1,44 @@
...
+            var searchButton = document.getElementById("search-button");
+            if (searchButton) {
+                searchButton.click();
+            }
```
MoralCode, line 39: Im not a fan of simulating a click on the button - would be better if we could call the actual search function or whatever the next stage of this process is.

MoralCode (changes requested):

EngCaioFonseca:
Closed due to #1112 #1113 and #1114

PR #1115, opened 2026-04-09
Remove need for rank from the visualizations

cdolfi:
This is an initial step to remove the need for rank which is computationally expensive on the materialized view level for minimal benefit 

### Generative AI disclosure

Please select one option:

- [ ] This contribution was NOT assisted or created by Generative AI tools.
- [x] This contribution was assisted or created by Generative AI tools.

If AI tools were used, please provide details below:
    - What tools were used?  claude
    - How were these tools used?  guidance on the transition 
    - Did you review these outputs before submitting this PR?  yes

Issue #1094, opened 2026-02-18
Are we upstreaming materialized views?


MoralCode:
Assuming that the heatmaps prs merge as intended

Now that we have split out materialized views into our own separate 8knot-only process, are we going to also upstream them into Augur? If we dont, someone who runs 8knot from a local augur instance wont have 2 of the necessary views.

Is that something we want to address?

cdolfi:
yes, I have periodically have been mentioning this

MoralCode:
another alternative is to explicitly create a process in 8knot that initializes any materialized views that don't exist/that arent reasonable to upstream, thats a little jank tho

MoralCode:
Views to upstream:

- repo files: https://github.com/oss-aspen/8Knot/blob/99f400e9e90d5323bffe48236bf73349218e8e53/docs/materialized_views/explorer_repo_files.sql#L4  (this apparrently was never in upstream CollectOSS)
- explorer_cntrb_per_file -> in infra repo
- explorer_contributor_actions -> infra repo
- explorer_pr_files -> infra repo


i think we also updated the definition for one view too

Edit: yep that was explorer_contributor_actions in the PR mentioned prior to this comment

MoralCode:
this is where these should go: https://github.com/chaoss/CollectOSS/blob/main/collectoss/application/db/materialized_views.py

PR #1116, opened 2026-04-09
Remove rank from all of the queries and update materialized view

cdolfi:
Rank is very costly without a lot of benefit #1115 removes the need for it and this PR is the 8knot updates needed when the materialized view is changed. NOTE: we will only want to merge this when we are making our prod change to the new materialized view

### Generative AI disclosure

Please select one option:

    - How were these tools used?  helped me optimize the query and build on different strategies
    - Did you review these outputs before submitting this PR?  yes, I made back and forth edits on the query

### DIFF
--- a/8Knot/cache_manager/db_init.py
+++ b/8Knot/cache_manager/db_init.py
@@ -204,7 +204,6 @@ def _create_application_tables() -> None:
                 repo_id int,
                 login text,
                 action text,
-                rank int,
                 cntrb_company text,
                 email_list text
             )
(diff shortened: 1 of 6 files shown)


MoralCode (changes requested):
code looks okay, going to ship this as a custom patch on prod.

marking as needs review to block merging it to 8knot until #1094 is complete (so that we dont create unnecessary complexity for newcomers)

MoralCode:
not merging this blocks our team from developing 8knot
merging it likely blocks newcomers in the same way per the above comment.

kind of rock and a hard place, but i think given we have conference demos coming up, we need to keep our ability to dev and test things

MoralCode (approved):

PR #1133, opened 2026-05-03
Add min_db: local CollectOSS-compatible postgres container for notebooks and dev

EngCaioFonseca:
This PR creates a local min db for development with a small data dump from Augur, so it is possible to set up and test it with small effort. 

### Generative AI disclosure

Please select one option:

    - How were these tools used?  Prototyping.


MoralCode (changes requested):
Per our maintainer sync:
- We should publish this as an image and add `image` to the service so users dont need to build it themselves, especially if it requires a SQL file they dont have
- can we name the folders/credentials something generic, like "sample_collected_data" and username/password?

docker/augur-db/init/04_rename_to_collection_data.sql
```diff
@@ -0,0 +1,10 @@
+-- min_db: align the loaded dump with the collectOSS schema.
+--
+-- The provided dump (03_data.sql) loads its data into the legacy `augur_data`
+-- schema. collectOSS-based 8Knot expects the schema to be named `collection_data`.
+-- Renaming is OID-safe: all tables, views, foreign keys, and materialized-view
+-- data follow the rename automatically, so no data is touched.
+--
+-- Runs after 03_data.sql due to alphabetical init ordering.
+
+ALTER SCHEMA augur_data RENAME TO collection_data;
```
MoralCode: Collrction_data is the old name

Its just data now

MoralCode (changes requested):
Since this is a one-off, we can probably store the setup instructions outside the repo if we need them again (I hope to replace this with an automatic job that keeps the container updated).

Thats fairly minor though.


Have we confirmed the image works for data science usecases (notebooks) too?

EngCaioFonseca:
Podman container up and built tested. 

Min_db working well.

Screenshot 2026-06-16 at 15 21 11

MoralCode:
`docker pull ghcr.io/oss-aspen/sample-collected-data:latest` published

PR #1151, opened 2026-05-13
Redis flush

### LINKED ISSUE #1150: Stale celery tasks in Redis are causing a deadlock after restart
the postgres cache gets initialized after restart while redis broker replays celery tasks from previous session.
One solution could be flushing redis at the start to get in sync with postgres.

EngCaioFonseca:
This PR resolves #1150. 

The pr flushes redis at start getting in sync with postgres start, since its cache starts fresh.

### Generative AI disclosure

Please select one option:

    - What tools were used?  Claude. 
    - How were these tools used?  Initial draft, brainstorming, prototype.

### DIFF
--- a/8Knot/cache_manager/db_init.py
+++ b/8Knot/cache_manager/db_init.py
@@ -61,7 +61,9 @@ Generally, 'int' is good for integers,
 
 import logging
 import sys
+import os
 import psycopg2 as pg
+import redis
 import time
 
 # doesn't use relative import syntax "import .cx_common" because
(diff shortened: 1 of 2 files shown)


8Knot/cache_manager/db_init.py
```diff
@@ -394,6 +396,35 @@ def _create_application_tables() -> None:
...
+    Flush the Redis broker so it stays in sync with postgres-cache.
+
+    postgres-cache uses UNLOGGED tables, so all cached data is lost on
+    restart. If Redis still holds stale Celery task messages or results
+    from a previous run, workers pick them up and poll for data that no
+    longer exists, causing deadlocks. Flushing here guarantees a clean
+    broker state every time the cache is (re)initialized.
+    """
+    broker_host = os.getenv("REDIS_SERVICE_HOST", "redis-broker")
+    broker_port = int(os.getenv("REDIS_SERVICE_PORT", "6379"))
```
MoralCode: This will crash if the environment variable is an empty string or other non-numeric value

MoralCode (dismissed):

8Knot/cache_manager/db_init.py
EngCaioFonseca: Resolved.

MoralCode: non-numeric values will still cause the new version of the code to fail (non empty strings are still truthy i think)

EngCaioFonseca: These are also handled now. Resolved.

cdolfi (approved):
lgtm

8Knot/cache_manager/db_init.py
MoralCode: the same feedback in my last comment still applies, except its now been moved into a new function

Why not use code like

```
    port_env = os.getenv("REDIS_SERVICE_PORT")
    broker_port = int(port_env) if port_env and port_env.is_numeric() else 6379
```

EngCaioFonseca: Your exception is already handled by the code:
"ABCD" → raw is truthy → int("ABCD") → ValueError → caught → returns 6379 and logs the misconfiguration. No crash.

_env_int() — a small helper that wraps the int() conversion in a try/except:

* If the env var is unset or empty→ returns the default (6379)
* If it's a valid number→ returns int(value)
* If it's a non-numeric string like"abc"→ catches ValueError, logs a warning so the misconfiguration is visible, and falls back to the default

MoralCode: ah ok yeah i missed the try catch mechanism. and i see that your function is able to log a warning (compared to mine)

lets make sure we file an issue/dont forget to do a quick check to see if theres anywhere else in the code where we are reading port numbers (and maybe take the opportunity to elevate this new function to a shared utility and turn the above comments into good quality docstrings)

MoralCode (commented):
havent personally run this branch, but submitting a comment to clear my "needs review" so this can merge on the basis of calis approval

8Knot/cache_manager/db_init.py
EngCaioFonseca: Just expanded the docstring of the function to contain these descriptions.

cdolfi (approved):
tested again locally and lgtm

PR #1153, opened 2026-05-13
Increase concurrency and change query order to reduce query bottleneck on local instances

### LINKED ISSUE #1152: Affiliation Query Bottleneck
Affiliation query is the most expensive SQL query operation and it's in the fifth position. We should rearrange the queries, and place it in last, since all the other ones finish way faster, and they wouldn't be blocked by it.
There's also just one worker going through each query one at a time, we can also scale up workers.

EngCaioFonseca:
This PR fixes #1152 

* The Workers-query concurrency is increased to parallelise the queries to avoid being deadlocked by slower and much complex queries like the affiliation q. 
So, the worker-query concurrency is bumped up to 8 to allow for 8 parallel queries to run, allowing for all faster queries to finish without the need to wait for affiliation q. Also this query is placed in the last place, so that it won't have any possibility of really blocking any query of being completed. Also, the sweet spot seems to be 8, as when the concurrency number is too high, there's too many long running queries hitting CollectOSS, which ends up dropping some of them and triggering a 3 min reset. 

* After increasing concurrency on the query's side, it is also necessary to bump up the callback concurrency, as the concurrency = 1, ends up blocking viz callbacks and freezing the page waiting for queries like affiliation to be completed. Once we increase the callback concurrency, any deadlock caused by complex queries is now resolved. Callbacks and viz can run without blocks, and pages get unfrozen, so the user can go to other pages if he wants as well without having to wait for the query or callback to finish. 

### Generative AI disclosure

Please select one option:

- [x] This contribution was NOT assisted or created by Generative AI tools.

    - What tools were used? 
    - How were these tools used? 

### DIFF
--- a/8Knot/pages/index/index_callbacks.py
+++ b/8Knot/pages/index/index_callbacks.py
@@ -37,7 +37,9 @@ from .search_utils import fuzzy_search
 from .search_utils import clean_repo_name
 
 # list of queries to be run (includes codebase page queries for heatmaps)
-QUERIES = [iq, cq, cnq, prq, aq, iaq, praq, prr, cpfq, rfq, prfq, rlq, pvq, rrq, osq, riq]
+# affiliation_query is last because its 3-way JOIN + GROUP BY is the slowest query;
+# keeping it at the end lets the faster queries finish without being blocked.
+QUERIES = [iq, cq, cnq, prq, iaq, praq, prr, cpfq, rfq, prfq, rlq, pvq, rrq, osq, riq, aq]
 
 
 # check if login has been enabled in config
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,7 +36,7 @@ services:
         ":8080",
         "app:server",
         "--workers",
-        "4",
+        "1",
         "--threads",
         "2",
         "--timeout",
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,7 +36,7 @@ services:
         ":8080",
         "app:server",
         "--workers",
-        "1",
+        "4",
         "--threads",
         "2",
         "--timeout",
@@ -67,7 +67,7 @@ services:
     build:
       context: .
       dockerfile: ./docker/Dockerfile
-    command: ["celery", "-A", "app:celery_app", "worker", "--loglevel=INFO", "--concurrency=1", "--time-limit=700", "--soft-time-limit=630"]
+    command: ["celery", "-A", "app:celery_app", "worker", "--loglevel=INFO", "--concurrency=8", "--time-limit=700", "--soft-time-limit=630"]
     depends_on:
       - redis-cache
       - redis-users
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -89,7 +89,7 @@ services:
         "--loglevel=INFO",
         "-Q",
         "data",
-        "--concurrency=1",
+        "--concurrency=2",
         "--time-limit=600",
         "--soft-time-limit=540"
       ]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -89,7 +89,7 @@ services:
         "--loglevel=INFO",
         "-Q",
         "data",
-        "--concurrency=2",
+        "--concurrency=8",
         "--time-limit=600",
         "--soft-time-limit=540"
       ]


EngCaioFonseca:
Benchmark test - Graphs loaded:
Dev branch with concurrency = 1 and affiliation query at  number 5 was 1 min and 20 s of loading time approximately.
Affiliation fix with concurrency = 8 and affiliation query at the last place -> 17s of loading time.

Both tests were performed after clearing everything, and the cache, through system prune and podman compose down --volumes.

MoralCode (commented):
Overall i agree, however this PR description needs WAY more context so that we can document the reasons behind it (especially since they were developed through in person conversation at Red Hat Summit 2026)

I was about to leave a comment asking for an explaination for why the `QUERIES` line was changed if all you were doing was adding a comment, until i saw the comment about how you decided to move one query to the end.

It takes a LOT of brain power to manually look through a diff like that to see what has changed. Having an explaination in the PR description of what the issue is, and what you have done to fix it would be helpful, especially for helping maintainers who werent part of the original conversations get caught up

MoralCode:
More details on your testing procedure would be helpful as well (rather than just the results) so that Cali and I can reproduce what you did

We discussed this over breakfast a little but ill put it in a comment here as well: I'd also like to maybe explore the concurrency a little more to understand why the system allows us to have 8 concurrent queries even without scaling the number of worker instances.

in other words, can we do some additional testing at different concurrency values to understand whether say, 4 or 16 concurrent queries makes the observed speedup more or less (indicating no bottleneck) or stays the same (indicating we aren't getting the full benefit of concurrency 8 and that something else, like the # of worker instances, may be the new bottleneck). Id like to better understand the relationship between celery scheduler concurrency and the number of workers (and especially make sure that this behavior/relationship gets documented)

EngCaioFonseca:
So, through my research, I was able to find why we can increase concurrency without scaling up workers. They are separate things that can be modified without necessarily affecting the other. 

Concurrency will allow multiple or paralleled tasks to be run at the same time.
The workers means compute instances, pods or containers. In the case of celery, it means a python process. 
So we can still have one worker, one containerized environment running parallel tasks, that is, running queries in parallel with just one worker. If we scale up the workers to two, that means two instances running the tasks in parallel. 

And yes, I'm still testing changing a bit the concurrency and scaling up workers to see if there's still improvement. But so far, the improvement in loading time and performance are very significant.

EngCaioFonseca:
Also, for all these related PRs I pushed recently, I've also created issues for all of them. Please check them as I've tried to describe in more detail the reason behind the PRs

MoralCode:
@EngCaioFonseca do you have links to the sources you found in your research? I'm very curious how the workers accomplish this parallelism. Are they just spinning up additional python processes? What are the limits for processes/compute in one worker before it makes sense to add another one? Etc

cdolfi:
This seems relatively straight forward to me. Will leave approval for @MoralCode once you get the context you are looking for

EngCaioFonseca:
@MoralCode one of the links that explains this:
https://khairi-brahmi.medium.com/mastering-celery-a-guide-to-background-tasks-workers-and-parallel-processing-in-python-eea575928c52
"The workers are simply background processes that listen for tasks in the queue and execute them when they appear."

but firstly, I got this explained to me by Claude as well, which was in a very concise and nice way to understand.

"Limits and the "when to add another worker" question
Resource cost of prefork:
Total resident memory is roughly (memory footprint of your app) × (concurrency + 1), where the +1 is the parent supervisor process. Each child also gets its own database connection, so with a concurrency of 10 you have 10 DB connections open. [Celery](https://celery.school/the-prefork-worker-pool)
CPU ceiling:
It's one CPU per child — a concurrency of 10 means you need 11 CPUs (10 children + 1 parent). The pool size is effectively capped by available CPUs. [Celery](https://celery.school/the-prefork-worker-pool)
When to scale out (add workers) vs. scale up (increase concurrency):
The practical heuristic:

CPU-bound tasks: max concurrency ≈ number of physical cores. Beyond that you're context-switching with no gain. Add machines (more workers on different hosts) to scale further.
I/O-bound tasks: switch to gevent/eventlet and you can push concurrency to hundreds on a single worker. Adding more workers then makes sense when you saturate the broker connection or need geographic/queue isolation.
Memory pressure: if your app is memory-hungry, you'll hit the RAM ceiling before the CPU ceiling. --autoscale=10,3 can help — it keeps a minimum of 3 processes alive but grows up to 10 when needed, and you can subclass Autoscaler to use custom metrics like load average or available memory. [Celery](https://docs.celeryq.dev/en/stable/userguide/workers.html)"

https://celery.school/the-prefork-worker-pool

The max concurrency or parallelized processes available are usually default to the number of CPU one has. 

Also, in terms of my own experimentation:
bumping up the concurrency to 16, will make the db drop the queries, as it can't handle this many queries, which then makes it worse as it would make it reset after 3 mins. Concurrency at 8, seems the sweet spot. 

Due to the affiliation query being a complex and slow query, even replacing it and increasing concurrency, it would a deadlock, blocking background callbacks and freezing page view as well, while they wait for the affiliation process to finish. The solution was to also increase concurrency of callback workers, this way, the page view/change and the other callbacks are unblocked through the parallelism of a higher callback concurrency.

MoralCode:
> Due to the affiliation query being a complex and slow query, even replacing it and increasing concurrency, it would a deadlock

What do you mean by this?

MoralCode:
the https://celery.school/the-prefork-worker-pool article talks about celery's default scheduling system and how it isn't great for IO bound tasks. While I think I generally support this change, I also think we should quickly look into what *is* recommended for heavy IO usecases (such as our query workers that are mostly just copying data from postgres). Maybe this is worth a new issue/followup PR

MoralCode (approved):
My questions are mostly just about trying to understand how the system works with the various settings available to us on the queries so we dont create any underlying things that are technically problems but have such a low/nil impact that they arent apparrent (until they maybe become part of a future bug chain)

overall though i think this is good. Struggling to test locally due to my podman bug, but it sounds like this has been well tested by caio and the diff is small enough that id be okay with the single person testing

cdolfi (changes requested):
this is what happens when I run locally, does not happen on dev: 

https://github.com/user-attachments/assets/4fb18512-de7a-4193-80ee-754ef0b4658e

cdolfi:
https://github.com/user-attachments/assets/768c5f00-408d-4a43-892e-45fbf8e6ceed

built right after a `podman system prune -af`

cdolfi (approved):
lgtm and tested on a fresh build

PR #1158, opened 2026-05-18
Create index for postgres cache tables

### LINKED ISSUE #1160: Index should be added into postgres-cache tables
Currently, retrieve_from_cache() does SELECT * WHERE repo_id IN (...) without an index. The UNLOGGED tables don't create index for themselves. This is quite substantial when talking about massive orgs/repos with millions of rows as it's a sequential scan instead of an indexed one.

EngCaioFonseca:
This PR's goal is to create index for every cache table. The reason why being that retrieve_from_cache() executes SELECT * ... without an index, which consequently means a sequential scan for potentially millions of rows. The UNLOGGED tables also don't auto-index. 
The impact, especially on prod and large repos, can be quite significant, in terms of complexity, it goes from O(n) to O(log n), so this could represent 10-100x faster on retrieval. 

### Generative AI disclosure

Please select one option:

    - How were these tools used?  Working on breaking down concepts and understanding layers of cache system. Initial draft.

### DIFF
--- a/8Knot/cache_manager/db_init.py
+++ b/8Knot/cache_manager/db_init.py
@@ -133,6 +133,42 @@ def _create_application_database() -> None:
     conn.close()
 
 
+_CACHE_TABLES_WITH_REPO_ID = (
+    "commits_query",
+    "issues_query",
+    "prs_query",
+    "affiliation_query",
+    "contributors_query",
+    "issue_assignee_query",
+    "pr_assignee_query",
+    "cntrb_per_file_query",
+    "pr_file_query",
+    "repo_files_query",
+    "repo_languages_query",
+    "package_version_query",
+    "repo_releases_query",
+    "ossf_score_query",
+    "repo_info_query",
+    "pr_response_query",
+)
+
+
+def _create_cache_indexes(cur) -> None:
+    """
+    Indexes for cache reads (retrieve_from_cache) and bookkeeping lookups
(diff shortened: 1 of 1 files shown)


8Knot/cache_manager/db_init.py
```diff
@@ -133,6 +133,42 @@ def _create_application_database() -> None:
...
+    "pr_assignee_query",
+    "cntrb_per_file_query",
+    "pr_file_query",
+    "repo_files_query",
+    "repo_languages_query",
+    "package_version_query",
+    "repo_releases_query",
+    "ossf_score_query",
+    "repo_info_query",
+    "pr_response_query",
```
MoralCode: i worry about listing this here - if we add new cache tables or change them, we have to remember to keep this updated.

Id rather convert the cache tables into SQLAlchemy syntax so we can define the indexes on them in the same way we do for CollectOSS

MoralCode (changes requested):

8Knot/cache_manager/db_init.py
EngCaioFonseca: Fair enough. Makes sense

MoralCode: what are we deciding to do about this feedback?

MoralCode:
gonna put this on hold until we implement schema versioning for the cache table - this would make a really good first migration. See: #1198

Issue #1155, opened 2026-05-14
decouple 8Knot from hard coded Schemas, enable user-defined schemas


MoralCode:
In an upcoming version, CollectOSS will be migrating its schema names from `augur_data` to something else.

As pointed out in #1119, there are many places in the code where we are hardcoding `augur_data` as a schema.

We should replace these hardcoded schema names with a search path.

I think this search path should be able to be set by an environment variable, but should have a default value such that it does not need to be set by every user who is using 8Knot in supported ways (such as with CollectOSS or legacy Augur instances).

I believe this can be done by setting the search path to `<new collectoss schema>,augur_data`. This should allow 8Knot to work with either schema name, preferring the new one, until we get to a point where we can fully deprecate the use of Legacy Augur.

The benefit of this is that we can (potentially, with lots of testing) avoid a hard cutoff that forces users to switch to CollectOSS before they use the next release of 8Knot.

MoralCode:
(feel free to file this comment as a new issue if desired)

I think alongside the new schema search path, we should add some kind of customizable string to 8Knot that displays an admin-configured name on the frontend to explain where the data came from. This would allow instance admins to state whether they are using 8Knot with CollectOSS, Augur, or other unsupported tools. Having this in the UI (and possibly also the logs) would help us better inform users about whether we can provide support for their issues based on the backend they are using.

This value could be set to "CollectOSS" (and maybe the collectoss db version number) if the default search path is detected, otherwise it could show either the user-defined string, or simply the user defined search path if a string is not available

PR #1163, opened 2026-05-18
Enable 8Knot to be flexible with what schema it uses

MoralCode:
work in progress PR for fixing #1155 and enabling 8Knot to keep working with data collection instances *both* before and after the CollectOSS fork.

This PR supersedes/replaces (and cherry picks small parts from) #1119 

Still need to:
- [ ] identify all the uses of hard-coded schemas in queries
- [ ] ensure the query paths for both the main DB and the cache DB are being set correctly (look for the new collectoss schema name first, then the legacy one)

### Generative AI disclosure

Please select one option:

- [X] This contribution was NOT assisted or created by Generative AI tools.


MoralCode:
since this changes queries, we should probably implement robust dataframe comparison testing first to be certain that it works

MoralCode:
superseded by #1176

PR #1176, opened 2026-05-29
Allow 8Knot to be compatible with both legacy and new CollectOSS database schemas

### LINKED ISSUE #1155: decouple 8Knot from hard coded Schemas, enable user-defined schemas

EngCaioFonseca:
This PR adapts 8knot for the new schema from CollectOSS by adding its schema name into the search path, as the first one to be used, and then defaults to augur_data if it doesn't work.

### Generative AI disclosure

Please select one option:

    - What tools were used?  Claude-Opus -4.8
    - How were these tools used?  Finding the places to fix for the new schema.

### DIFF
--- a/8Knot/cache_manager/cx_common.py
+++ b/8Knot/cache_manager/cx_common.py
@@ -13,10 +13,11 @@ try:
     env_augur_host = os.environ["AUGUR_HOST"]
     env_augur_port = os.environ["AUGUR_PORT"]
     env_augur_database = os.environ["AUGUR_DATABASE"]
-    env_augur_schema = os.environ["AUGUR_SCHEMA"]
 except KeyError as ke:
     logging.critical(f"AUGUR: Database credentials incomplete: {ke}")
     raise KeyError(ke)
+
+env_augur_schema = os.getenv("AUGUR_SCHEMA", "collection_data,augur_data")
 
 # credentials to access application cache from environment
 env_dbname = os.getenv("CACHE_DB_NAME", "augur_cache")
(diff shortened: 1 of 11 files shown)


MoralCode (changes requested):
Sorry to do this to you, but the new name has changed from `collection_data` to just `data`. Can you update this PR?

8Knot/cache_manager/cx_common.py
```diff
@@ -26,7 +27,7 @@
 env_user = os.getenv("CACHE_USER", "postgres")
 env_password = os.getenv("POSTGRES_PASSWORD", "password")
 env_port = os.getenv("CACHE_PORT", "5432")
-env_schema = os.getenv("CACHE_SCHEMA", "augur_data")
+env_schema = os.getenv("CACHE_SCHEMA", "data,augur_data")
```
MoralCode, line 30: Is 8knot changing its cache schema names too?

EngCaioFonseca: should we, or should we not? I was just working on changing or adapting everything to it

MoralCode: This is an internal implementation detail that should only matter within 8knot. Also the cache gets erased regularly, so if this isnt already part of this PR, we can probably address it in a follow-up PR later

This change might be something thats worth doing as part of whatever PR series we do to rename 8knot internal stuff to remove old naming for the CollectOSS db.

MoralCode (approved):
Tested this with databases that used each different schema (prod read - old schema, and my local dev - with new schema and the materialized views that are needed)

i was using the stock compose file with minor irrelevant tweaks. No configured cache volume, so nothing that needed to be reset run to run.

Either way, 8Knot seems to function about the same on either database. Id say this is good to merge

Issue #1070, opened 2026-01-19
Make it where it is no longer needs to manually reset the cache for new data


cdolfi:
Right now once data is queried for a specific repo new data will not be shown unless the 8knot cache is manually reset. 

A potential route for this would be to check the last_collected_on data for the related tables and if that date is after the last time the query was done then the query is done again an cached. 

Very open to discussion of how this is handled

MoralCode:
> if that date is after the last time the query was done

Are we already storing this somewhere?

Also it would probably help to know more about the cache architecture - are we caching per-query (i.e. cache will be stored differently for a request to analyze chaoss/augur and foo/bar vs a request for foo/bar and bar/baz? or is it stored per-repo/group, causing the date for foo/bar to be reused?)

cdolfi:
Currently we are not, something we will need to think about how to do it. Open to other ideas on implementation, but think we should move from admins needed to manually reset the cache asap 

we cache by repo to make sure there is not repeat caching for different sets

MoralCode:
Im curious about the scope of this issue. Are we looking for a band-aid (automating the cache resetting) or an actual fix?


Either are fine as long as we make a new issue for the actual fix if we choose to band aid now.

also i think a complicating factor is that our instance is architected fundamentally differently than most contributors are using augur, so which usecase is our primary concern with this issue? (It sounds like we want to make it easier for the community since we already clear the cache internally in a systematic way)



Some ideas i have that i think are interesting for us (collcetively) explore as we roadmap the idela future of 8knot architecture:
- what if we bypass the cache entirely? Our judgement of the need for a cache is a) unique to our instance and hosting, and b) based on an old hosting architecture we dont use from 9 months ago. Most contributors will run 8knot side by side with a tiny augur instance on their machines probably. At that scale, a cache seems like more trouble than its worth or could be moved to a "special considerations for production deployment" doc 
- Can we move some or all of our cache to a materialized view? 
- Josh said we have potentially a lot to gain by not running two instances of postgres. If we still need A cache, can we:
    - move the cache to use a new schema in our same 8knot instance, reducing duplicate postgres instances
    - move the cache to something in-memory like redis or an in memory SQLite DB.
- Can we speed up our queries (through direct optimization of how we query and through postgres config) enough that we dont need a cache, even for maybe simple production deployments contributors may make at medium scale? How would this change for our instance?

[...]

PR #1177, opened 2026-06-04
Shareable short url

### LINKED ISSUE #706: Make graphs / pages shareable via URL
**Describe the bug**
I can't find a way to share a specific page of graphs with other people - Ideally there'd be a way to copy/paste the url to a specific graph that preserves my current view options, but it'd also be helpful to share the default results for a specific repo/dataset.

**To Reproduce**
1. Load up the data for a specific repository
2. Copy the URL
3. Paste the URL into a new anonymous browser window

**Expected behavior**
The new browser window should have the same view as the original one

EngCaioFonseca:
This PR implements a shareable URL through a 2 layer architecture. We're able to encapsulate as much info into the url to allowing for exact matching when sharing, but also through a URL shortener, share a compact URL for a better user experience. 
This PR resolves #706 

The system design/architecture is as follows:

Shareable URL - 2 Layers(1)

Screenshot 2026-06-04 at 16 09 59

### Generative AI disclosure

Please select one option:

    - What tools were used?  Claude - Opus 4.8, Coderabbit, Claude - Fable 5.
    - How were these tools used?  Initial draft, prototype, tests.
    - Did you review these outputs before submitting this PR?  Yes, went back and forth on the design/architecture, implementation, evaluation, etc.


8Knot/cache_manager/db_init.py
```diff
@@ -421,6 +421,25 @@ def _create_application_tables() -> None:
...
+                full_state    TEXT         NOT NULL,
+                created_at    TIMESTAMPTZ  NOT NULL DEFAULT NOW(),
+                last_accessed TIMESTAMPTZ,
+                access_count  INTEGER      NOT NULL DEFAULT 0,
+                expires_at    TIMESTAMPTZ  NOT NULL DEFAULT (NOW() + INTERVAL '90 days')
+            )
+            """
+        )
+        cur.execute("CREATE INDEX IF NOT EXISTS share_links_expires_idx ON share_links (expires_at)")
+        logging.warning("CREATED share_links TABLE")
```
MoralCode: you are creating the share_links table in db-init? isnt that the process that initializes the cache table (i.e. the table that we regularly reset and want to automate the resetting of in #1070? )

If we regularly delete this table, we will regularly be breaking urls that people share, and the experience will be decidedly worse (because we will be promising a stable feature and then breaking that stability)

MoralCode (changes requested):

8Knot/cache_manager/db_init.py
EngCaioFonseca: yes, that makes sense. I'll remove it from the db-init and place it on the appropriate place.

EngCaioFonseca:
This PR will now implement only the first layer of the shareable URL, without the URL shortener. 
Having it's main goal, to be able to share URLs that will lead to the exact page and graph view the user is seeing and sharing. 

This PR fixes a previous issue where the graph view was not right when trying to share URLs, the page would be not centred on the right/specific graph. 

Cause and fix for the issue: The end of the URL would have, or would try to get the viz id for the anchor to work and go towards that view. However, there was a mismatch between the URL element id and how we are defining the element through visualizationAIO. In the VizAIO the element or the graph id is defined as the:

```
id = f"{page}-{viz_id}"     # e.g. "affiliation" + "-" + "commit-domains"
```

But the share button only knew the graph's viz_id, so the first version of the share URL ended in:

```
#commit-domains    ← no element on the page has this id.
```

so, fixing the URL DOMs id solves the problem. 


The layer 1 of this architecture is as follows:

Shareable_URL_layer1

A summary of the workings could be described as:

* Share packs the selection + graph into a ?state= blob and appends the card's real DOM id (#{page}-{viz_id}). 
* Open decodes the blob, re-runs the normal search, so the graph renders
* Then a client-side loop polls for the card and re-centers through the page's async reflow - landing you exactly on the shared graph. 
* On this layer of the implementation, no need for dbs or persistent storage anywhere, as we're just enconding the URL straight from the app.

EngCaioFonseca:
High-level explanation of the code (pseudocode):

## On the Sharing side - manage_share_modal -> _build_share_url
** share_callbacks.py
Once the share button is clicked:
    triggered -> carries {graph: viz_id, page: page}
    selection -> projects.value, that's the search bar pills (orgs + repos)

    if there's no selection: 
          show "run a search first" then stop/

* the selection is then split through _split_selection:
repo_ids, org_names

* then for the encoding step through _encode_state (Layer 1 -> the whole share)
** cache_manager/url_state.py 
payload -> version number, repo_ids, org_names, pathname, graph_id, filters 
blob -> base64url ( gzip (json(payload))) -> approximately 200 chars

* The DOM id anchor -> with the graph view fix for this PR
** share_callbacks.py
will have the anchor be -> page + viz_id  #card's real id
 
* The composition of the URL will then be:
    url = base + pathname + "?state=" + blob + "#" + anchor

## on the Consuming (opening the link side) -> handle_share_url -> apply_share_state
** share_callbacks.py
on url.search changes (fires on page load as well):
   blob = extract "state" param
   if there's no blob: stop.   # normal navigation, nothing is being shared.

* decoding step through decoding_state
** cache_manager/url_state.py 
if the len of blob > 16 KB: reject
bytes = bounded_gunzip (base_64url_decode(blob) ) # zip-bomb guard
payload = json (bytes)
payload = migrate(payload) if in the old version # forward compat seam
if not shape_valid(payload): reject

* Validation
** pages/utils/graph_registry.py
if not validate_target(pathname, graph_id): warn "graph removed"; stop

* resolve selection through the _resolve_selection

## Landing on the exact graph
** share_callbacks.py
on url.hash or share-loaded-state changes:
 id = sanitize(hash)     # e.g. "affiliation-unique-domains"
cancelled = false

on the first user wheel/touch -> canceled = true  # so it never fights the user

function scroll():
if cancelled: done
else = getElementById(id)
if else: el.scrollIntoView(center); return FOUND
else: return NOT_YET

cdolfi:
@EngCaioFonseca We can work through this more when I get back, most important to find a format that works for you and @MoralCode . I would say even more high level. My initial interpretation would me something like: 

To share: 
- encoding approach 
    - a bullet point or two summarizing approach 
- The composition of the URL: url = base + pathname + "?state=" + blob + "#" + anchor

To open the link: 
- a bullet point or two / sub bullet point(s) on how to go from link to inital set up 
- how to get to graph bullet point 
   - a sub point or two maybe 
- how to get search option bullet point 
   - a sub point or two maybe

PR #1186, opened 2026-06-09
add .gitattributes to fix python line endings

### LINKED ISSUE #1062: Add .gitattributes for defining how to handle line endings

Ishant5436:
Hey, saw this good first issue and wanted to help out. Added a `.gitattributes` file to enforce LF for Python scripts and prevent the CRLF rewriting issues mentioned in #1061. Let me know if you need any changes!


Ishant5436:
closing

PR #1194, opened 2026-06-19
release version displayed in the bottom left corner

EngCaioFonseca:
This PR will display the latest release version on the bottom left corner. 

### Generative AI disclosure

Please select one option:

    - How were these tools used?  Initial draft.
    - Did you review these outputs before submitting this PR?  Yes, reviewed, fixed it, and tested.


docker/Dockerfile
```diff
@@ -10,13 +10,14 @@ COPY pyproject.toml pyproject.toml
...
 RUN uv sync --python=/opt/app-root/bin/python --locked --no-install-project --no-dev --no-cache
 
-# remove uv binary copied in earlier to hopefully save some space
+COPY ./8Knot/ /opt/app-root/src/
+
+# Bake the release version from pyproject.toml into the app, then remove build files
 USER root
-RUN rm /bin/uv pyproject.toml uv.lock
+RUN sed -nE 's/^version *= *"([^"]+)".*/\1/p' pyproject.toml > /opt/app-root/src/VERSION \
+    && rm /bin/uv pyproject.toml uv.lock
```
MoralCode: This feels like an odd way to do it when the version information is already in pyproject.toml

docker-compose.yml
```diff
@@ -52,6 +52,8 @@ services:
     #   - 8080:8080
     volumes:
       - ./8Knot:/opt/app-root/src/:z
+      # source mount above shadows the image's baked VERSION; expose pyproject.toml so the version still resolves
+      - ./pyproject.toml:/opt/app-root/pyproject.toml:ro,z
```
MoralCode: pyproject.toml should already be in `/opt/app-root/src/`. is that not the case?

8Knot/pages/index/search_utils.py
```diff
@@ -26,6 +26,8 @@
 import dash_bootstrap_components as dbc
 import dash_mantine_components as dmc
 
+from _version import get_version
```
MoralCode: what file is this importing from? Gemini is telling me that this is not a python builtin or anything and requires a `_version.py` file to exist in the project.

i wonder if this would be a better way: https://stackoverflow.com/a/75100875/

MoralCode (changes requested):
Had some thoughts on the code, curious what you think of my alternative way to do it

docker-compose.yml
EngCaioFonseca: The Dockerfile copies it to /opt/app-root/ (not src/) and then deletes it; only ./8Knot goes to src/. And in dev, ./8Knot is bind-mounted over src/, which is what shadowed the baked file.

8Knot/pages/index/search_utils.py
EngCaioFonseca: I actually didn't add the version.py, I'm doing it now.

EngCaioFonseca:
check this new version @MoralCode, which auto-generates the version through the pre-commit importing it from pyproject.toml, since it seems it's not possible to directly get it on runtime.

EngCaioFonseca:
working now with the metadata file. 

Screenshot 2026-07-06 at 00 31 41

pyproject.toml
```diff
@@ -23,3 +27,12 @@ dependencies = [
     "redis~=6.4",
     "sqlalchemy~=2.0",
 ]
+
+# The release version lives in 8Knot/metadata.py so the running app can import it
+# (pyproject.toml is not shipped into the container). Same pattern as CollectOSS.
+[tool.setuptools]
+package-dir = { "" = "8Knot" }
+packages = []
+
```
MoralCode, line 36: This part doesnt seem to be present in https://github.com/chaoss/CollectOSS/blob/adfa6e89b95f9c43f5093e72a4f4ac95a884b03b/pyproject.toml#L118-L124

Do you know why its needed?

8Knot/pages/index/search_utils.py
```diff
@@ -643,8 +645,8 @@ def create_bottom_navbar():
 
     return dbc.NavbarSimple(
         children=children,
-        brand="",
-        brand_href="#",
+        brand=html.Span(f"v{__version__}", className="version-display"),
+        brand_href=None,
```
MoralCode, line 649: Minor thing for maybe a future version:

We should maybe find a way to indicate if the container was modified/built from source. Maybe this needs to be done after we release official image builds so we can add a `(release)` flag in this UI on release versions. (this maybe could be its own issue)

MoralCode (commented):
Overall LGTM - tested and works.

Just had one small comment about a difference in how this code works vs how CollectOSS does it (also probably related to why 8knot needs to depend on itself in the lockfile)

pyproject.toml
EngCaioFonseca: it's needed due to their location on the repo tree. CollectOSS keeps metadata.py at the root, and so only needed  packages.find to narrow down the search for it. In 8Knot this lives in 8Knot/metadata.py, so the setuptools has to be pointed to the directory.  
For 8Knot, if the metadata.py was in the repo-level, it would be invisible to the running app. Which was the reason we're not going with reading the pyproject.toml at runtime.

pyproject.toml
```diff
@@ -23,3 +27,12 @@ dependencies = [
     "redis~=6.4",
     "sqlalchemy~=2.0",
 ]
+
+# The release version lives in 8Knot/metadata.py so the running app can import it
+# (pyproject.toml is not shipped into the container). Same pattern as CollectOSS.
```
MoralCode: i dont think this comment line
`# (pyproject.toml is not shipped into the container). Same pattern as CollectOSS.`
is correct

MoralCode (changes requested):

MoralCode (approved):

PR #1196, opened 2026-07-06
Shareable URL with the right anchor set to each graph view

EngCaioFonseca:
This PR implements a shareable URL button, where the user can share the url for each graph view on the app. Previous anchor was implemented using only the page view, instead of the right anchor id, which is set as viz + page_view id. 

 

### Generative AI disclosure

Please select one option:

    - What tools were used?  Claude Fable 5
    - How were these tools used?  Draft, review.
    - Did you review these outputs before submitting this PR?  Yes, reviewed and tested.

### DIFF
--- a/8Knot/pages/index/share_callbacks.py
+++ b/8Knot/pages/index/share_callbacks.py
@@ -1,0 +1,266 @@
+"""
+Callbacks for the shareable URL system.
+
+The whole share is carried in the URL: the searchbar selection + target graph
+are serialized into a versioned gzip+base64url payload under ``?state=``. No
+database — the link is self-contained and survives any cache reset. (A
+DB-backed short-URL layer, ``?s=<id>``, is a planned follow-up PR.)
+
+Integration rules (each one prevents a class of Dash breakage):
+  * Share UI lives in `share_components.create_share_modal()`, NOT in the
+    app stores list.
+  * The load path never writes `repo-choices` directly — it populates
+    `projects` and bumps `search-button.n_clicks`, reusing the existing
+    search pipeline (no `allow_duplicate` races).
+  * Only ONE callback reads `url.search`.
+  * Pattern-matching share-button IDs are callback INPUTS only; every
+    OUTPUT targets a simple string-id modal component.
+"""
+
+from __future__ import annotations
+
+import dash
+from dash import callback, ctx, clientside_callback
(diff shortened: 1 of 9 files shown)


MoralCode:
Can you include more details in the PR description for how this works? like is this relying on specific/documented dash features to do the "jump to"?

I like the end behavior of this (it even updated the scroll position when a graph above it loaded and got taller and shifted the layout!), but the size of this diff makes it hard to be thorough with the code review step in the time window i have available. 

Not every day that the PR is tested and working before code review happens but i guess such is the strange timeline of software we are now in

8Knot/pages/utils/graph_registry.py
```diff
@@ -0,0 +1,86 @@
...
+from __future__ import annotations
+
+VALID_GRAPH_REGISTRY: dict[str, list[str]] = {
+    "/repo_overview": [
+        "code-languages",
+        "ossf-scorecard",
+        "package-version",
+        "repo-general-info",
+    ],
+    "/contributions": [
```
MoralCode, line 20: We can do this dynamically in the VisualizationAIO - which has access to the ID being given to it for nav purposes

MoralCode: updating the share logic to use the `            id=(self.nav_id if self.nav_id is not None else f"{self.page}-{self.viz_id}"),` logic from vizaio should eliminate the need for this registry

8Knot/pages/index/share_callbacks.py
MoralCode, line 1: if this wont affect the "jump to graph" can we move the state things to their own PR?

MoralCode (changes requested):

EngCaioFonseca:
Okay, taking a deeper look into the workings of the anchor fix in this branch vs how it works on prod, is that the real DOM id is indeed necessary, but not sufficient. When we test on prod a url like: ...affiliation#affiliation-gh-org-affiliation, it fails because of when the browser scrolls: the native hash-scrolling happens once, at document load - at that moment Dash hasn't client-side-rendered the page yet, so the element doesn't exist. Even though it is the correct ID, there's nothing to scroll to, silent no-op.

It works on this branch, because the clientside scroll handler is keyed on url.hash, independent of the ?state=, so once the card appears, it goes on to center it. 

for the v1, we don't need the state, like I mentioned, just the handler.

PR #1197, opened 2026-07-10
fix Shareable URL fragment scrolling bug

EngCaioFonseca:
This PR implements the anchor fix and url handler to correctly center the page on the graph being shared.

Two fixes make URLs land centered on the target graph:

1. Sidebar anchors now use each card's real DOM id, f"{PAGE}-{VIZ_ID}" — the old fragments matched no element, and PAGE is not derivable from the pathname (e.g. /repo_overview cards are repo_info-*). Clicking a sidebar graph now also puts a copyable, shareable anchor URL in the address bar.

2. A clientside handler scrolls to the fragment. The browser's native hash scroll fires before Dash client-side-renders the page (why a correct anchor still does nothing in prod), so the handler polls for the card, centers it, re-scrolls through the async graph-load reflow, and cancels if the user scrolls.

### Generative AI disclosure

Please select one option:

    - What tools were used?  Claude Fable
    - How were these tools used?  Draft and review.
    - Did you review these outputs before submitting this PR?  Yes, reviewed and tested.

### DIFF
--- a/8Knot/assets/scroll_to_anchor.js
+++ b/8Knot/assets/scroll_to_anchor.js
@@ -1,0 +1,45 @@
+/*
+ * Scroll the URL's #fragment target into view. Registered as a Dash clientside
+ * callback on url.hash — see index_callbacks.py.
+ *
+ * The target ids already exist: each graph card carries a nav id (the
+ * VisualizationAIO `id=` set in the viz files), and the codebase heatmaps use
+ * the ids on their layout rows. This adds NO ids — it only fixes WHEN the
+ * scroll happens:
+ *   - Dash renders the page client-side AFTER the browser's one native hash
+ *     scroll, so at that moment the target element does not exist yet.
+ *   - Graphs above the target then load asynchronously and reflow the page,
+ *     moving the target. That settle time is variable and unbounded (cached
+ *     vs live data, count/kind of graphs above, network), so no single delay
+ *     is correct — too short misses the slow tail, too long stalls every page.
+ *
+ * So we keep the target pinned at center on a short interval until the layout
+ * settles: each tick is a no-op before the element exists and a no-op once it
+ * is centered, and it self-corrects instantly as the page reflows. It stops at
+ * the first user scroll (armed only after our first scroll, so leftover scroll
+ * momentum can't cancel it early) or after a 6s cap.
+ */
+window.dash_clientside = Object.assign({}, window.dash_clientside, {
+    eightknot: {
+        scroll_to_anchor: function (hash_) {
+            if (!hash_) { return window.dash_clientside.no_update; }
+            var id = hash_.replace(/^#/, "");
+            var stopAt = Date.now() + 6000;
+            var armed = false;
+            var timer = setInterval(function () {
+                var el = document.getElementById(id);
+                if (el) {
+                    el.scrollIntoView({ block: "center", behavior: "instant" });
+                    if (!armed) {
+                        armed = true;
+                        var cancel = function () { clearInterval(timer); };
+                        window.addEventListener("wheel", cancel, { passive: true, once: true });
+                        window.addEventListener("touchmove", cancel, { passive: true, once: true });
+                    }
+                }
+                if (Date.now() > stopAt) { clearInterval(timer); }
+            }, 200);
+            return window.dash_clientside.no_update;
+        },
+    },
+});
--- a/8Knot/pages/index/index_callbacks.py
+++ b/8Knot/pages/index/index_callbacks.py
@@ -7,7 +7,7 @@ import json
 from celery.result import AsyncResult
 import dash_bootstrap_components as dbc
 import dash
-from dash import callback, html
+from dash import callback, html, clientside_callback, ClientsideFunction
 from dash.dependencies import Input, Output, State, MATCH
 from app import augur
 from flask_login import current_user
@@ -897,3 +897,13 @@ def update_pill_color_on_search(search_button_clicks, selected_repos_orgs):
         return "searchbar-dropdown"
 
     return dash.no_update
+
+
+# Scroll the URL's #fragment graph card into view on every hash change.
+# Implementation (and the reasoning for it) lives in assets/scroll_to_anchor.js.
+clientside_callback(
+    ClientsideFunction(namespace="eightknot", function_name="scroll_to_anchor"),
+    Output("anchor-scroll-dummy", "data"),
+    Input("url", "hash"),
+    prevent_initial_call=False,
+)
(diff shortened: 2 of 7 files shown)


EngCaioFonseca:
We can also change the modifications on the NavAIO ids, and leave it as it was, and just change the card graph IDs to match the nav ids. We could do it either way, but personally, I think changing the nav ids to match the graph ids makes more sense.

8Knot/pages/index/index_components.py
MoralCode, line 1: changing the navigation links like this wasnt part of my understanding of the plan - also the navbar links already work correctly

8Knot/pages/index/index_callbacks.py
```diff
@@ -897,3 +897,56 @@ def update_pill_color_on_search(search_button_clicks, selected_repos_orgs):
...
+# Fires on every hash change, so sidebar graph links scroll correctly too.
+# ============================================================================
+clientside_callback(
+    """
+    function(hash_) {
+        if (!hash_) { return window.dash_clientside.no_update; }
+        var id;
+        try { id = decodeURIComponent(hash_.replace(/^#/, '')); }
+        catch (e) { return window.dash_clientside.no_update; }
+        if (!id) { return window.dash_clientside.no_update; }
```
MoralCode: Asked this on slack, but i can see why i missed this function the first time

Can you walk me through the simpler versions of this code that you tried before arriving on this as the solution? 

Id love to figure out how we can make this more readable/maintainable for the future (and any potential contributors that want to learn how it works and why.


The part im most skeptical of is the repeatedly checking for the elements at arbitrary time intervals - how do we know the page wont take more than 5.5 seconds to show the DOM elements? Is there anything else that this polling behavior can get in the way of? Would it be better to find a way to ensure the graphs take up just as much space when theyre added to the DOM as they do when theyre fully loaded so that we dont need to account for any shifts in content?

MoralCode (changes requested):

8Knot/pages/index/index_components.py
EngCaioFonseca: discussed through coworking meeting. Review comment verified.

8Knot/assets/scroll_to_anchor.js
```diff
@@ -0,0 +1,48 @@
...
+                var el = document.getElementById(id);
+                if (!el) { return false; }
+                var r = el.getBoundingClientRect();
+                // only scroll if the card is meaningfully off-center
+                if (Math.abs(r.top + r.height / 2 - window.innerHeight / 2) > 40) {
+                    el.scrollIntoView({ behavior: "smooth", block: "center" });
+                }
+                return true;
+            }
+            var tries = 0;
```
MoralCode: This is not what we discussed

MoralCode (changes requested):

8Knot/assets/scroll_to_anchor.js
```diff
@@ -0,0 +1,43 @@
...
+            var timer = setInterval(function () {
+                var el = document.getElementById(id);
+                if (el) {
+                    el.scrollIntoView({ block: "center", behavior: "instant" });
+                    if (!armed) {
+                        armed = true;
+                        var cancel = function () { clearInterval(timer); };
+                        window.addEventListener("wheel", cancel, { passive: true, once: true });
+                        window.addEventListener("touchmove", cancel, { passive: true, once: true });
+                    }
```
MoralCode, line 38: we've already established through previous testing that simply waiting for the target scroll element to be present in the DOM doesnt work as the trigger for a single scroll action

8Knot/components/visualization.py
```diff
@@ -90,7 +92,9 @@ def __init__(
                 ),
             ],
             className=class_name,
-            id=(self.nav_id if self.nav_id is not None else f"{self.page}-{self.viz_id}"),
+            # Card DOM id doubles as the URL anchor (#<id>); defaults to the
+            # bare viz_id so anchors stay readable and defined in one place.
+            id=(self.nav_id if self.nav_id is not None else self.viz_id),
```
MoralCode: this has fundamentally changed the existing behavior of the scroll anchors and is - in my opinion - not necessary

8Knot/assets/scroll_to_anchor.js
EngCaioFonseca: Agreed on the single-scroll-on presence dfails, but that's exactly why this isn't a single scroll. The interval will re-center continuously through the reflow, presence-check only no-ops the early ticks. 
It calls scrollintoview on every tick the element exists, so it will be recorrect as the page reflows.

8Knot/components/visualization.py
EngCaioFonseca: but this was discussed earlier. We have 3 options basically, using the one-line logic for id; remap the sidebar fragments, which was the option I went for previously (33 hrefs in one file), or explicit id= on each card, which would be modifying 33 viz files. 
So, this one line logic was what we agreed on earlier.
A few things on it:
The card's default DOM id changed from {page}-{viz-id}-> viz_id.
there was no working scroll-anchor behavior to break. on prod the sidebar ids never matched the card ids, so the anchors didn't scroll at all, which is the reason this would solve it.
It also fixed a duplicate id issue we discussed before. Before, the card and its dcc.graph both had id {page}-{viz-id}. Now card=viz-id, graph={page}-{viz-id}, so no collision.

In the end, the whole issue is that the anchor #X can only scroll to the element with the id="X". The sidebar ids are bare names; the card ids were prefixed. So they must match somewhere, which is why we come to the 3 options above. We agreed then to go on with the one line logic.

MoralCode: > there was no working scroll-anchor behavior to break. on prod the sidebar ids never matched the card ids, so the anchors didn't scroll at all,

This is incorrect

the sidebar nav relies on fragment scrolling using the built-in browser behavior.

All the graph cards (except the heatmap ones) already have an assigned ID specified for use in navigation scrolling that was added at the time we refactored the visualizations to AIO. The graphs also had a similar ID that matches the {page}-{viz-id} format, which may be the source of the confusion

EngCaioFonseca: okay, noted. I'll modify it once again to use those ids.

EngCaioFonseca:
https://github.com/user-attachments/assets/76bae1e9-0b00-4391-bcff-cee6f4d23dba

MoralCode (approved):
Im not a fan of the complexity of the clientside callback code (looping and repeatedly trying to scroll) but we've spent enough time trying other things so I think this is probably fine as a "working enough" solution for now

Issue #358, opened 2023-06-07
Callback rendering clogged if the user switches pages.


JamesKunstle:
If a user triggers graphs to render but then changes pages, both sets of visualizations render. This means that the second set- the graphs that the user is waiting for- are being slowed down by the previous page's graphs, even though their result is going to be thrown away.

Investigate how we can cancel all of the graphs rendering for a user when a page switch occurs. The buttons that change pages may allow middleware or something.

Long callbacks are cancelable so we just need to call them in that respect: 
https://dash.plotly.com/long-callbacks

JamesKunstle:
This is being worked on by the Plotly team as a new feature:

https://github.com/plotly/dash/issues/2588

Recent PR adds callbacks that trigger cancellation on page refresh. Upcoming work could optimize this by having Dash frontend trigger cancellation of known outdated long callbacks.

MoralCode:
the underlying dash-layer  issue of auto-cancelling callbacks seems to not be fully implemented yet

I think doing this ourselves with the long callbacks is our best short term bet (if we can find a dash input that changes when the page is about to be closed).

Long term i think we should try and resume the conversations about upstreaming this if dash wants to support it (seems like the thread was mostly being driven by James)

MoralCode:
from #1201 

> could maybe have a time running after a tab is closed and then if not reopened, then all queries are eliminated. I think it's an option.

if the tab is closed, no client side code runs. where do you propose we put the code?

MoralCode:
if a user closes a tab by mistake, theyll open a new tab and trigger fresh queries, so we dont want queries from their last tab to remain open (as they will delay loading the new queries)

PR #1201, opened 2026-08-14
Cancel DB queries when the 8Knot instance shuts down

### LINKED ISSUE #802: Cancel queries once app is brought down
There are long running queries against the DB that are not getting ended even once the 8knot instance is shut off

EngCaioFonseca:

There seems to have 4 factors contributing to this issue: 

Closes [#802](https://github.com/oss-aspen/8Knot/issues/802)

Long-running queries against the Augur database kept running after the 8Knot instance was shut down. Four things combined to cause this:

No statement_timeout on any Augur connection, so Postgres had no reason to ever stop.
Postgres does not check whether the client is still there while it is busy inside a query — it only finds out on the next write to the socket. A container killed mid-SELECT left the server burning CPU indefinitely.
with psycopg2.connect(...) does not close the connection; it only ends the transaction. A worker killed between fetches left its session idle in transaction, holding the server-side cursor open.
Celery's --soft-time-limit cannot interrupt a blocking psycopg2 execute(), so the limits we already had, bounded the worker, not the query. That is what produced the orphans in the first place.

The fix is on the server-side, where no client-side handler gets to run.

Every connection to Augur now carries statement_timeout, idle_in_transaction_session_timeout, and TCP keepalives (set on the server's end of the socket). On PostgreSQL 14+ we also enable client_connection_check_interval, which makes Postgres poll for a vanished client during a query and abort promptly. That setting also requires POLLRDHUP, which is Linux-only, so it is applied best-effort — hosts that reject it log a warning and stay bounded by the timeouts.

Two supporting changes:

Query tasks no longer retry on QueryCanceled. Adding statement_timeout means a slow query now fails instead of hanging, and the existing autoretry_for=(Exception,) with max_retries: 5 would have retried it five times, each burning another 500s of Augur DB time. A dont_autoretry_for on the base task class prevents that.
caching_wrapper re-raises the original exception instead of flattening it into a generic Exception, which is what lets the above actually match.

Three tunables were added to .env.sample, all in milliseconds: AUGUR_STATEMENT_TIMEOUT_MS (500s, worker queries — kept under Celery's 540s soft limit so the DB aborts first), AUGUR_ENGINE_STATEMENT_TIMEOUT_MS (30min, the app-server's searchbar query, which is legitimately much slower), and AUGUR_IDLE_TX_TIMEOUT_MS (2min).

### Generative AI disclosure

Please select one option:

    - What tools were used?  Claude Opus 5, GPT 5.6 Sol
    - How were these tools used?  Investigation, initial draft, review.
    - Did you review these outputs before submitting this PR?  Yes, reviewed it.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -4,6 +4,11 @@ AUGUR_PASSWORD=augur
 AUGUR_PORT=5432
 AUGUR_SCHEMA=data,augur_data
 AUGUR_USERNAME=augur
+# Bound how long augur keeps working on our behalf, so queries don't outlive a
+# shut-down 8Knot instance. Milliseconds.
+AUGUR_STATEMENT_TIMEOUT_MS=500000
+AUGUR_ENGINE_STATEMENT_TIMEOUT_MS=1800000
+AUGUR_IDLE_TX_TIMEOUT_MS=120000
 
 DEBUG_8KNOT=True
 REDIS_PASSWORD=1234
--- a/.env.sample
+++ b/.env.sample
@@ -4,8 +4,8 @@ AUGUR_PASSWORD=augur
 AUGUR_PORT=5432
 AUGUR_SCHEMA=data,augur_data
 AUGUR_USERNAME=augur
-# Bound how long augur keeps working on our behalf, so queries don't outlive a
-# shut-down 8Knot instance. Milliseconds.
+# Bound individual source DB statements and abandoned transactions so they do
+# not continue indefinitely after worker or app-server shutdown. Milliseconds.
 AUGUR_STATEMENT_TIMEOUT_MS=500000
 AUGUR_ENGINE_STATEMENT_TIMEOUT_MS=1800000
 AUGUR_IDLE_TX_TIMEOUT_MS=120000
--- a/8Knot/_celery.py
+++ b/8Knot/_celery.py
@@ -1,5 +1,6 @@
-from celery import Celery
+from celery import Celery, Task
 from dash import CeleryManager
+from psycopg2.errors import QueryCanceled
 import os
 
 redis_host = "{}".format(os.getenv("REDIS_SERVICE_HOST", "redis-broker"))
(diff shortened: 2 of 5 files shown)


cdolfi:
Tested this out locally. I opened up 8knot and searched for a few repos, looked at the running queries on the db and then closed the tab. When I checked the queries after I closed it I still saw some of the queries still running from my now closed 8knot window

MoralCode:
> then closed the tab. 

The scope of this PR by my understanding is canceling queries when the instance shuts down. Its very confusingly similar language to the original goal of "when a user closes the tab". Maybe its a phase 1 or something. I'm not sure if I fully remember the context or how this scope changed but the linked issue seems to have a title that reflects this "whole instance shutdown" scope too

EngCaioFonseca:
Yes, @MoralCode is right. So the goal of the PR and the issue so far is to be able to eliminate the persistent queries even after app/instance shutdown. Which it's working on this PR. Closing the window/tab of 8Knot and being able to kill the queries could be a next step, although open for discussion as well, because, do we really want to kill the queries just after a tab closes? If a user closes it accidentally, everything would have to start again, in terms of queries. We could maybe have a time running after a tab is closed and then if not reopened, then all queries are eliminated. I think it's an option.
But for this PR, I'd say, having the time limits, and being able to validate that no query will remain open after the instance is shutdown, already solves in a significant way this issue with the queries.

MoralCode:
> validate that no query will remain open after the instance is shutdown, already solves in a significant way this issue with the queries.

I disagree. We are only shutting 8Knot down in prod once a week (and thats because we have a special process, i imagine other instances shut down less often than that). The main argument i see for this could be for development systems.

If queries are still active in postgres when 8knot disappears, that sounds like a potential postgres configuration issue

cdolfi:
After testing current dev, queries are already shutting down when the app is being brought down. So we do not need change anything for that functionality. I agree with adrain as the issue of users closing tabs but the queries running for long periods was the problem that #802 was attempting to describe

MoralCode:
> I agree with adrain

That wasnt what i was saying. IMO #802 is a potentially valid (if lacking in detail/evidence) independent issue report that is clearly talking about instance shutdown. IMO we should put any evidence of queries staying open after instance shutdown there (or close it due to lack of reproducibility), and maybe shelve it until that issue becomes more pressing than whats in our current roadmap. I dont want to lose the research that has happened so far, but this is a fairly small performance fish. 

IMO conversation about canceling no-longer-needed queries on tab closure or user navigation should go in #358

EngCaioFonseca:
Yes, I agree. I'm going to keep this PR open dealing with persistent queries after app is shutdown, and do more investigation on the #358 to check if we can reproduce it or if that's still happening, then we, potentially (that's the goal), already have the solution to that in this PR.  

For the issue we were discussing, being able to eliminate persistent queries after the tab/window is closed, I'm working on another branch in order to deal with that.

PR #1189, opened 2026-06-12
Add label filter to Issues Over Time graph

Dipro-cyber:
### Pull Request Change Description

Adds an "Issue Filter" control to the Issues Over Time visualization, allowing users to toggle between viewing all issues or only those tagged with a "bug" label.

**Changes:**
- Added a `RadioItems` control ("All Issues" / "Bugs Only") to the visualization controls
- Wired it into the callback as a new `Input`
- Filters the DataFrame in `process_data()` on the `labels` column when "Bugs Only" is selected

**Note:** This relies on the `labels` column and `issue_labels` JOIN added to `issues_query` in PR #[your #1085 PR number]. Both PRs should be reviewed together or this one merged after that one.

Closes #254

### Generative AI disclosure
- [x] This contribution was assisted or created by Generative AI tools.
  - What tools were used? Kiro (AI-powered dev environment)
  - How were these tools used? Code generation following existing patterns in the codebase
  - Did you review these outputs before submitting this PR? Yes

PR #1204, opened 2026-08-25
Adding cache schema version

### LINKED ISSUE #1198: Version 8Knot cache schema
- it needs to check its version, and check if there’s new data in, and if needs to be upgraded to the most up to date version.

Steps:

- Check/learn how DB versioning works in collectoss
- apply this versioning to 8Knot’s cache schema

EngCaioFonseca:

This solves #1198 and adds the cache schema versioning as well as lay the groundwork for the follow up steps or making the cache persistent.

The cache schema is now versioned with Alembic. It runs on startup: a fresh cache gets stamped at the latest version, and an existing cache gets any missing migrations applied automatically. Base tables are still created with the raw CREATE blocks in db_init.py, so adding a new table works exactly like before. Alembic only handles changes to existing tables. 

The first migration adds the labels column from #1189. Existing caches never got it, because CREATE TABLE IF NOT EXISTS doesn't touch a table that already exists. The migration adds the column and clears the old cached issues so they get re-collected with labels.

A few things added for the cache long-term:

- build the repo_id indexes without locking the tables (CREATE INDEX CONCURRENTLY)
- only flush the Redis broker when the cache actually changed instead of on every startup, and stopped it from also wiping user sessions
- take a lock during init so two app pods starting at once don't step on each other
- build the DB connection strings with psycopg2's make_dsn so passwords with special characters don't break them

This is also groundwork for making the cache persistent instead of rebuilt on every boot.

### Generative AI disclosure

Please select one option:

    - What tools were used?  Claude - Opus 4.8; GPT - 5.6 Sol
    - How were these tools used?  Draft, review iteration.

### DIFF
--- a/8Knot/cache_manager/db_init.py
+++ b/8Knot/cache_manager/db_init.py
@@ -11,9 +11,9 @@ create a table in the cache. Most people who would be working on a
 project like this will know enough SQL to read the existing table
 definitions and create a new table as needed from those examples.
 
-Our data model is fairly simple, so for now the overhead of proper
-db migration tooling is mostly bloaty. We can return to this decision
-in the future if necessary.
+Our data model is fairly simple, so we do not use alembic. Schema
+changes to existing tables are numbered SQL functions in MIGRATIONS
+below, applied on startup. We can return to alembic later if needed.
 """
 
 """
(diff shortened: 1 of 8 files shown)


MoralCode:
#1189 made changes to the schema, so this may need to change to account for that

8Knot/cache_manager/db_init.py
```diff
@@ -11,9 +11,9 @@
 project like this will know enough SQL to read the existing table
 definitions and create a new table as needed from those examples.
 
-Our data model is fairly simple, so for now the overhead of proper
-db migration tooling is mostly bloaty. We can return to this decision
-in the future if necessary.
+Our data model is fairly simple, so we do not use alembic. Schema
+changes to existing tables are numbered SQL functions in MIGRATIONS
+below, applied on startup. We can return to alembic later if needed.
```
MoralCode: can you elaborate on why we aren't using alembic? This seems like a very odd technical decision to me, especially if we are using SQL functions in its place

EngCaioFonseca: Fair point, and I kinda agree, but in this we're not removing alembic, or touching it in this repo. I've updated the comment to explain the reasoning behind it. In summary, since it is a small fixed schema, as well as disposable/unlogged and rebuildable cache, it's easier and more efficient to not use it and it's the reason why we don't need it.

8Knot/cache_manager/migrations/versions/0001_add_labels_to_issues_query.py
```diff
@@ -0,0 +1,33 @@
+"""add labels column to issues_query
+
+Adds the labels column to caches created before it (CREATE TABLE
```
cdolfi: Is this suppose to be an example? a bit confused on the scope of this file to this overall PR

PR #1209, opened 2026-09-04
Fix/cache empty repolist error handling

EngCaioFonseca:

Fixes four issues in PostgreSQL data-access and startup error handling. The changes are local and introduce no schema, dependency, or callback-graph changes.

### 1. Handle empty repository selections

`repo-choices` initially contains `[]`, but its callbacks still run. Psycopg2 converted the empty sequence into `IN ()`, causing invalid SQL during initial page loads.

- `get_uncached()` now returns `[]` immediately.
- `retrieve_from_cache()` uses `IN (NULL)` to return zero rows while preserving DataFrame columns.
- Heatmap and repository-overview callbacks now return cleared or no-data states when no repository is selected.
- `get_default_repo_with_data()` now returns `None` instead of indexing an empty list.

This is handled at the data and callback boundaries rather than with `prevent_initial_call`, which would break navigation with selections restored from session storage.

### 2. Preserve database error causes

`AugurManager.run_query()` previously caught every exception and discarded its cause.

It now:

- Catches only `Exception`, allowing `KeyboardInterrupt` and `SystemExit` to propagate.
- Logs the full traceback.
- Chains the original error with `raise ... from e`.

### 3. Preserve `_bots.py` startup failures

If `AugurManager()` failed, `_bots.py` continued with an uninitialized `dbm`, masking the original error with `NameError`.

The original `KeyError` or `SQLAlchemyError` is now logged and re-raised.

### 4. Correct cache bookkeeping

`caching_wrapper()` recorded every requested repository, including repositories already cached. This created duplicate bookkeeping rows.

Query parameters and uncached repository IDs now use separate variables, and bookkeeping records only repositories fetched by the current operation.

## Result

- Empty repository selections no longer cause SQL or callback errors.
- Database failures retain their original cause and traceback.
- Startup failures are no longer masked.
- Cache bookkeeping no longer adds duplicates for already-cached repositories.

## The added testing will check for:

- Empty repository selections do not trigger invalid queries or callback exceptions.
- Empty query results retain the columns that downstream processing expects.
- Cache bookkeeping records only newly fetched repositories.
- Database and startup failures preserve their original causes.
- Dropdowns can be cleared and selected again safely.

### Generative AI disclosure

Please select one option:

    - What tools were used?  Claude Opus 5, GPT 5.6 - sol, GPT 6 - Astra.
    - How were these tools used?  Draft, multiple iterations, review.


cdolfi:
Do we have any examples of these errors happening? What prompted the making of this PR?

